PR #1, opened 2017-02-03
[HYDRATOR-585] Implementation for http sink plugin

shashankmoghe:
Design: https://wiki.cask.co/display/CE/HTTPSink
JIRA: https://issues.cask.co/browse/HYDRATOR-585


docs/HTTP-batchsink.md
```diff
@@ -0,0 +1,61 @@
...
+
+
+Description
+-----------
+Sink plugin to send the messages from the pipeline to an external http endpoint.
+
+Properties
+----------
+
+**url:** The URL to post data to.
```
russorat: which ones are macro enabled?

docs/HTTP-batchsink.md
```diff
@@ -0,0 +1,61 @@
...
+
+**method:** The HTTP request method. Defaults to POST.
+
+**bathSize:** Batch size. Defaults to 1.
```
russorat: bath?

docs/HTTP-batchsink.md
```diff
@@ -0,0 +1,61 @@
...
+
+**messageFormat:** Format to send messsage in. Options are JSON,Form,Custom. Defaults to JSON.
```
russorat: space after each comma

docs/HTTP-batchsink.md
```diff
@@ -0,0 +1,61 @@
...
+
+**body:** Custom message. This would be mandatory if message format is selected as Custom.
```
russorat: "This is required if the message format is set to 'Custom'"

docs/HTTP-batchsink.md
```diff
@@ -0,0 +1,61 @@
...
+          In case user wants to have Custom message and want to use some of the input record fields as variables to build the message,
```
russorat: this is worded very poorly. How does a user add fields from the message to the custom post payload? can you also add an example?

src/main/java/co/cask/hydrator/plugin/batch/HTTPSink.java
```diff
@@ -0,0 +1,402 @@
...
+import javax.net.ssl.X509TrustManager;
+import javax.ws.rs.HttpMethod;
+
+/**
+ * Sink plugin to send the messages from the pipeline to an external http endpoint.
+ */
+@Plugin(type = BatchSink.PLUGIN_TYPE)
+@Name("HTTP")
+@Description("Sink plugin to send the messages from the pipeline to an external http endpoint.")
+
```
russorat: remove empty line

src/main/java/co/cask/hydrator/plugin/batch/HTTPSink.java
```diff
@@ -0,0 +1,402 @@
...
+    if (config.method.equals("POST") || config.method.equals("PUT")) {
+      if (config.messageFormat.equals("JSON")) {
+        message = StructuredRecordStringConverter.toJsonString(input);
+        contentType = "application/json";
+      } else if (config.messageFormat.equals("Form")) {
+        message = createFormMessage(input);
+        contentType = " application/x-www-form-urlencoded";
+      } else if (config.messageFormat.equals("Custom")) {
+        message = createCustomMessage(config.body, input);
+        contentType = "application/plain";
```
russorat: application/plain is not a common content type according to http://www.freeformatter.com/mime-types-list.html

we probably shouldn't set this anyway if it's custom.

src/main/java/co/cask/hydrator/plugin/batch/HTTPSink.java
```diff
@@ -0,0 +1,402 @@
...
+        HttpURLConnection conn = null;
+        Map<String, String> headers = config.getRequestHeadersMap();
+        try {
+          URL url = new URL(config.url);
+          conn = (HttpURLConnection) url.openConnection();
+          if (conn instanceof HttpsURLConnection) {
+            //Disable SSLv3
+            System.setProperty("https.protocols", "TLSv1,TLSv1.1,TLSv1.2");
+            if (config.disableSSLValidation) {
+              disableSSLValidation();
```
russorat: should this be an option, similar to the httptohdfs action?

src/main/java/co/cask/hydrator/plugin/batch/HTTPSink.java
```diff
@@ -0,0 +1,402 @@
...
+          conn.setRequestMethod(config.method.toUpperCase());
+          conn.setConnectTimeout(config.connectTimeout);
+          conn.setReadTimeout(config.readTimeout);
+          conn.setInstanceFollowRedirects(config.followRedirects);
+          conn.addRequestProperty("charset", config.charset);
+          for (Map.Entry<String, String> propertyEntry : headers.entrySet()) {
+            conn.addRequestProperty(propertyEntry.getKey(), propertyEntry.getValue());
+          }
+          if (config.method.equals("POST") || config.method.equals("PUT")) {
+            if (!headers.containsKey("Content-Type")) {
```
russorat: we should overwrite the one we set if the user wants to use a different one.

src/main/java/co/cask/hydrator/plugin/batch/HTTPSink.java
```diff
@@ -0,0 +1,402 @@
...
+              conn.addRequestProperty("Content-Type", contentType);
+            }
+          }
+
+          if (messages.length() > 0) {
+            conn.setDoOutput(true);
+            try (OutputStream outputStream = conn.getOutputStream()) {
+              LOG.info("Payload: " + messages.toString().trim());
```
russorat: remove info log

src/main/java/co/cask/hydrator/plugin/batch/HTTPSink.java
```diff
@@ -0,0 +1,402 @@
...
+              outputStream.write(messages.toString().trim().getBytes(config.charset));
+            }
+          }
+          responseCode = conn.getResponseCode();
+          messages.setLength(0);
+          if (config.failOnNon200Response && responseCode != 200) {
```
russorat: this needs to be any 200 level response, so >= 200 && < 300

src/main/java/co/cask/hydrator/plugin/batch/HTTPSink.java
```diff
@@ -0,0 +1,402 @@
...
+    boolean first = true;
+    String formMessage = null;
+    StringBuilder sb = new StringBuilder("");
+    for (Schema.Field field : input.getSchema().getFields()) {
+      if (first) {
+        first = false;
+      } else {
+        sb.append("&");
+      }
+      sb.append(field.getName() + "=" + input.get(field.getName()));
```
russorat: do these values need to be URL encoded?

russorat: also, this should be 3 sb.append statements

src/main/java/co/cask/hydrator/plugin/batch/HTTPSink.java
```diff
@@ -0,0 +1,402 @@
...
+      formMessage = URLEncoder.encode(sb.toString(), config.charset);
+    } catch (UnsupportedEncodingException e) {
+      throw new IllegalStateException("Error encoding Form message. Reason: " + e.getMessage(), e);
+    }
+    return formMessage;
+  }
+
+  private String createCustomMessage(String body, StructuredRecord input) {
+    String customMessage = body;
+    String regexForHashedVariables = "#s*(\\w+)";
```
russorat: should this be a static constant?

src/main/java/co/cask/hydrator/plugin/batch/HTTPSink.java
```diff
@@ -0,0 +1,402 @@
...
+   * Config for the HTTP sink.
+   */
+  public static class HTTPSinkConfig extends ReferencePluginConfig {
+
+    @Description("The URL to post data to.")
+    @Macro
+    private String url;
+
+    @Description("The http request method. Defaults to POST")
+    private String method;
```
russorat: macro

src/main/java/co/cask/hydrator/plugin/batch/HTTPSink.java
```diff
@@ -0,0 +1,402 @@
...
+
+    @Description("Batch size. Defaults to 1.")
+    @Macro
+    private Integer batchSize;
+
+    @Nullable
+    @Description("Delimiter for messages to be used while batching. Defaults to \"\\n\"")
```
russorat: macro

src/main/java/co/cask/hydrator/plugin/batch/HTTPSink.java
```diff
@@ -0,0 +1,402 @@
...
+    private String delimiterForMessages;
+
+    @Description("Format to send messsage in.")
+    private String messageFormat;
```
russorat: macro

src/main/java/co/cask/hydrator/plugin/batch/HTTPSink.java
```diff
@@ -0,0 +1,402 @@
...
+    @Macro
+    private String body;
+
+    @Nullable
+    @Description("Request headers to set when performing the http request.")
+    @Macro
+    private String requestHeaders;
+
+    @Description("Charset. Defaults to UTF-8.")
+    private String charset;
```
russorat: macro

src/main/java/co/cask/hydrator/plugin/batch/HTTPSink.java
```diff
@@ -0,0 +1,402 @@
...
+
+    @Description("Whether to automatically follow redirects. Defaults to true.")
+    private Boolean followRedirects;
```
russorat: macro

src/main/java/co/cask/hydrator/plugin/batch/HTTPSink.java
```diff
@@ -0,0 +1,402 @@
...
+
+    @Description("If user enables SSL validation, they will be expected to add the certificate to the trustStore" +
+      " on each machine. Defaults to true.")
+    private Boolean disableSSLValidation;
```
russorat: macro

src/main/java/co/cask/hydrator/plugin/batch/HTTPSink.java
```diff
@@ -0,0 +1,402 @@
...
+
+    @Description("The number of times the request should be retried if the request fails. Defaults to 3.")
+    private Integer numRetries;
```
russorat: macro

src/main/java/co/cask/hydrator/plugin/batch/HTTPSink.java
```diff
@@ -0,0 +1,402 @@
...
+    @Macro
+    private Integer connectTimeout;
+
+    @Description("The time in milliseconds to wait for a read. Set to 0 for infinite. Defaults to 60000 (1 minute).")
+    @Nullable
+    @Macro
+    private Integer readTimeout;
+
+    @Description("Whether to fail the pipeline on non-200 response from the http end point. Defaults to true.")
+    private Boolean failOnNon200Response;
```
russorat: macro

src/main/java/co/cask/hydrator/plugin/batch/HTTPSink.java
```diff
@@ -0,0 +1,402 @@
...
+          if (conn != null) {
+            conn.disconnect();
+          }
+        }
+        retries++;
+      } while (retries < config.numRetries);
+      if (exception != null) {
+        throw exception;
+      }
+    }
```
russorat: how do you handle tokens left over at the end. for example, if batch size is 100, and you get 101 messages, what happens to the last message?

src/main/java/co/cask/hydrator/plugin/batch/HTTPSink.java
shashankmoghe: yes,this would be an option using config.disableSSLValidation

src/main/java/co/cask/hydrator/plugin/batch/HTTPSink.java
shashankmoghe: Please confirm.
If failOnNon200Response is true and response code is in the range of 201 to 300,we need to throw the exception.

src/main/java/co/cask/hydrator/plugin/batch/HTTPSink.java
shashankmoghe: Thank you for this.I will check and update.

shashankmoghe: Please suggest one below:
Option1:We can override destroy() and if 'messages' is not empty then send the http request with the remaining messages.
Option2:We can take input from user on total number of records and use in transform() to send the http request for remaining records/messages.

src/main/java/co/cask/hydrator/plugin/batch/HTTPSink.java
russorat: @shashankmoghe no. Any 2XX response code is successful. so if failOnNon200Response = true, and a 201 response is returned, the pipeline should NOT throw an exception.

src/main/java/co/cask/hydrator/plugin/batch/HTTPSink.java
shashankmoghe: If user has entered Content-Type in the headers then we are adding the same else the default one is being set for each message format.

src/main/java/co/cask/hydrator/plugin/batch/HTTPSink.java
shashankmoghe: In case user doesnt add a header property for content type,we would need this as "text/plain" for custom.

src/main/java/co/cask/hydrator/plugin/batch/HTTPSink.java
shashankmoghe: Implemented the Option1.Please review

russorat:
LGTM

PR #10, opened 2019-07-10
CDAP-15524. HTTP Batch Source. Plugin code. Mapreduce routine. Configs.

aonischuk:
Includes:

1. Generic plugin implementation.
2. Implementation of MR routine.
3. Plugin config and related classes. Including configs validation.

For the sake of easy review. PR does not include everything implemented so far. That's why you see
"Not yet implemented" in StringToRecordConverterFactory.java and BaseHttpPaginationIterator.java.
These classes are bridge to another part of functionality, which will be presented in a next PR.


docs/HTTP-batchsource.md
```diff
@@ -388,5 +388,6 @@ error.
 **Transport Protocols:** Transport protocols which are allowed for connection.
 
 **Cipher Suites:** Cipher suites which are allowed for connection.
+Colons, commas or spaces are also acceptable separators.
```
albertshau, line 391: the fact that this is a delimited list is hidden from UI users. We shouldn't mention this and should use a delimiter where this isn't a problem.

We really should add support for maps and lists in CDAP...

src/main/java/io/cdap/plugin/http/source/batch/HttpBatchSource.java
```diff
@@ -0,0 +1,146 @@
...
+
+    pipelineConfigurer.getStageConfigurer().setOutputSchema(config.getSchema());
+  }
+
+  @Override
+  public void prepareRun(BatchSourceContext context) {
+    config.validate(); // validate when macros are already substituted
+    config.validateSchema();
+
+    if (schema == null) {
```
albertshau: Unless I'm missing something, schema is always null at this point. Doesn't seem like this if is needed

src/main/java/io/cdap/plugin/http/source/batch/HttpBatchSource.java
```diff
@@ -0,0 +1,146 @@
...
+        .collect(Collectors.toList()));
+
+    context.setInput(Input.of(config.referenceName, new HttpInputFormatProvider(config)));
+  }
+
+  private BaseStringToRecordConverter stringToRecordConverter;
+
+  @Override
+  public void initialize(BatchRuntimeContext context) throws Exception {
+    this.schema = context.getOutputSchema();
```
albertshau: this will be whatever was set at configure time. So if schema is a macro, this will be null. It should always be correct to get this from the plugin config instead.

src/main/java/io/cdap/plugin/http/source/batch/HttpBatchSource.java
```diff
@@ -0,0 +1,146 @@
...
+
+    super.initialize(context);
+  }
+
+  @Override
+  public void transform(KeyValue<LongWritable, Text> input,
+                        Emitter<StructuredRecord> emitter) throws IOException {
+    String resultString = input.getValue().toString();
+    try {
+      Integer httpCode = new Long(input.getKey().get()).intValue();
```
albertshau: if this is expected to be an int, the InputFormat should just use an Integer

src/main/java/io/cdap/plugin/http/source/batch/HttpBatchSource.java
```diff
@@ -0,0 +1,146 @@
...
+    this.stringToRecordConverter = StringToRecordConverterFactory.createInstance(config, schema);
+    this.httpErrorHandler = new HttpErrorHandler(config);
+
+    super.initialize(context);
+  }
+
+  @Override
+  public void transform(KeyValue<LongWritable, Text> input,
+                        Emitter<StructuredRecord> emitter) throws IOException {
+    String resultString = input.getValue().toString();
```
albertshau, line 109: For the blob format that will need to be supported, it will be incorrect to represent the response as a string. I think it's better to have the InputFormat return StructuredRecords, with the record conversion logic in the InputFormat instead of in this transform method.

src/main/java/io/cdap/plugin/http/source/batch/HttpInputFormatProvider.java
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * InputFormatProvider used by cdap to provide configurations to mapreduce job
+ */
+public class HttpInputFormatProvider implements InputFormatProvider {
+  private final Map<String, String> conf;
+
+  HttpInputFormatProvider(HttpBatchSourceConfig config) {
+    this.conf = new MapreduceConfigBuilder<String, String>()
+      .putIfValueNotNull(BaseHttpSourceConfig.PROPERTY_REFERENCE_NAME, config.referenceName)
```
albertshau: it's probably easier to use Gson to serialize the entire config as a json string and have just a single property instead of having all of these separate properties. If we add any more properties, it will be very difficult to tell if they've all been correctly added here.
I think that would also allow us to remove those massive constructors.

src/main/java/io/cdap/plugin/http/source/batch/HttpRecordReader.java
```diff
@@ -0,0 +1,89 @@
...
+import java.io.IOException;
+
+/**
+ * RecordReader implementation, which reads text records representations and http codes
+ * using {@link BaseHttpPaginationIterator} subclasses.
+ */
+public class HttpRecordReader extends RecordReader<LongWritable, Text> {
+
+  private static final Logger LOG = LoggerFactory.getLogger(HttpRecordReader.class);
+  private BaseHttpPaginationIterator httpPagionationIterator;
```
albertshau: typo: Pagionation -> Pagination

src/main/java/io/cdap/plugin/http/source/batch/HttpRecordReader.java
```diff
@@ -0,0 +1,89 @@
...
+
+  @Override
+  public Text getCurrentValue() {
+    return value;
+  }
+
+  @Override
+  public float getProgress() {
+    // progress is unknown
+    return 1.0f;
```
albertshau: should probably use 0 instead of 1. Showing no progress is better than showing full progress.

src/main/java/io/cdap/plugin/http/source/common/BaseHttpSourceConfig.java
```diff
@@ -0,0 +1,867 @@
...
+
+  @Nullable
+  public Map<String, String> getHeadersMap() {
+    return ConfigUtil.getMapFromKeyValueString(headers);
+  }
+
+  @Nonnull
+  public Map<Pattern, HttpErrorHandlingStrategy> getHttpErrorsHandlingMap() {
+    Map<String, String> httpErrorsHandlingMap = ConfigUtil.getMapFromKeyValueString(httpErrorsHandling);
+    Map<Pattern, HttpErrorHandlingStrategy> result = new LinkedHashMap<>(); // order of pattern checks is important
```
albertshau: since order is important, it's better to return a List of objects instead of a Map so that the contract is more clear.  Introduce a new class that is just a pair of Pattern and HttpErrorHandlingStrategy.

In general, if we only ever iterate through Map entries and never get by the key, it's a sign that a Map isn't the right data structure.

src/main/java/io/cdap/plugin/http/source/common/BaseHttpSourceConfig.java
```diff
@@ -0,0 +1,867 @@
...
+
+    for (Map.Entry<String, String> entry : httpErrorsHandlingMap.entrySet()) {
+      result.put(Pattern.compile(entry.getKey()),
```
albertshau: if the pattern is invalid, we should throw an exception with a nice message.

src/main/java/io/cdap/plugin/http/source/common/ConfigUtil.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.Map;
+import java.util.stream.Stream;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+/**
+ * Methods for configs validation and configs conversion.
+ */
+public class ConfigUtil {
+  public static void assertIsSet(Object propertyValue, String propertyName, String reason) {
```
albertshau: are these methods going to be used anywhere except the config? If not it's better just to put them in the config class.

src/main/java/io/cdap/plugin/http/source/common/ConfigUtil.java
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+
+  public static <T extends EnumWithValue> T
+  getEnumValueByString(Class<T> enumClass, String stringValue, String propertyName) {
+    return Stream.of(enumClass.getEnumConstants())
+      .filter(keyType -> keyType.getValue().equalsIgnoreCase(stringValue))
+      .findAny()
+      .orElseThrow(() -> new InvalidConfigPropertyException(
+        String.format("Unsupported '%s' property value: '%s'", propertyName, stringValue), propertyName));
```
albertshau: Unsupported value for '%s': '%s'

src/main/java/io/cdap/plugin/http/source/common/ConfigUtil.java
```diff
@@ -0,0 +1,95 @@
...
+  @Nullable
+  public static Long toLong(String value, String propertyName) {
+    if (Strings.isNullOrEmpty(value)) {
+      return null;
+    }
+
+    try {
+      return Long.parseLong(value);
+    } catch (NumberFormatException ex) {
+      throw new InvalidConfigPropertyException(String.format("Unsupported value for %s: %s", propertyName, value),
```
albertshau: Unsupported value for '%s': '%s'

src/main/java/io/cdap/plugin/http/source/common/ConfigUtil.java
```diff
@@ -0,0 +1,95 @@
...
+                                               propertyName);
+    }
+  }
+
+  @Nonnull
```
albertshau: We can leave this out. By convention, anything in our projects that is not annotated as Nullable is non-null.

src/main/java/io/cdap/plugin/http/source/common/HttpErrorHandler.java
```diff
@@ -0,0 +1,45 @@
...
+    String httpCodeString = Integer.toString(httpCode);
+
+    for (Map.Entry<Pattern, HttpErrorHandlingStrategy> entry : httpErrorsHandlingMap.entrySet()) {
+      Matcher matcher = entry.getKey().matcher(httpCodeString);
+      if (matcher.matches()) {
+        return entry.getValue();
+      }
+    }
+
+    throw new IllegalStateException(String.format("No error handling strategy defined for HTTP status code '%d'. " +
```
albertshau: Think it's better to have a default instead of requiring that every status code possibility is covered. Perhaps failing should be the default?

In both cases the pipeline will fail, but we want it to fail with a message about the HTTP error, not with this error.

src/main/java/io/cdap/plugin/http/source/common/HttpErrorHandlingStrategy.java
```diff
@@ -0,0 +1,62 @@
...
+  @Override
+  public String toString() {
+    return this.getValue();
+  }
+
+  public boolean shouldRetry() {
+    return (this.equals(RETRY_AND_FAIL) || this.equals(RETRY_AND_SKIP) || this.equals(RETRY_AND_SEND_TO_ERROR));
+  }
+
+  public HttpErrorHandlingStrategy getPostRetryStrategy() {
```
albertshau: I'm a bit confused about what 'PostRetry' means (first time I saw it above I thought it was for 'POST' requests).

src/main/java/io/cdap/plugin/http/source/common/HttpErrorHandlingStrategy.java
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Indicates error handling strategy which will be used to handle unexpected http status codes.
+ */
+public enum HttpErrorHandlingStrategy implements EnumWithValue {
+  SUCCESS("Success"),
+  FAIL("Fail"),
+  SKIP("Skip"),
+  SEND_TO_ERROR("Send to error"),
+  RETRY_AND_FAIL("Retry and fail"),
```
albertshau: Now that I think of this more, I don't think we need the 'retry and xyz' strategies. Retries and failure handling can be orthogonal. We can just have Fail, Skip, Send to Error. If no retry strategy is given, no retries will be done. Otherwise the error strategy only applies after retries have been exhausted.

src/main/java/io/cdap/plugin/http/source/common/HttpErrorHandlingStrategy.java
```diff
@@ -0,0 +1,62 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.cdap.plugin.http.source.common;
+
+/**
+ * Indicates error handling strategy which will be used to handle unexpected http status codes.
+ */
+public enum HttpErrorHandlingStrategy implements EnumWithValue {
+  SUCCESS("Success"),
```
albertshau: what does success mean?

src/main/java/io/cdap/plugin/http/source/common/pagination/BaseHttpPaginationIterator.java
```diff
@@ -0,0 +1,31 @@
...
+import org.apache.commons.lang3.tuple.ImmutablePair;
+
+import java.io.Closeable;
+import java.util.Iterator;
+
+/**
+ * An iterator which iterates over every element on all the pages of the given resource. The page urls are generated
+ * based on configured {@link PaginationType}.
+ */
+public abstract class BaseHttpPaginationIterator implements Iterator<ImmutablePair<Integer, String>>, Closeable {
```
albertshau: In most cases (including this one), I think it's better to create a Java class with two fields rather than use Pair classes. It makes it more clear what the elements represent.

docs/HTTP-batchsource.md
aonischuk: Actually I made it a plain text area. Not a list. Why? Cipher suites are usually copy pasted and can contain 30-40 properties. It would be very un-usable to add every of them manually to list in UI manually.

Example of cipher suites config:
`TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA384,TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA384,TLS_RSA_WITH_AES_256_CBC_SHA256,TLS_ECDH_ECDSA_WITH_AES_256_CBC_SHA384,TLS_ECDH_RSA_WITH_AES_256_CBC_SHA384,TLS_DHE_RSA_WITH_AES_256_CBC_SHA256,TLS_DHE_DSS_WITH_AES_256_CBC_SHA256,TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA,TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA,TLS_RSA_WITH_AES_256_CBC_SHA,TLS_ECDH_ECDSA_WITH_AES_256_CBC_SHA,TLS_ECDH_RSA_WITH_AES_256_CBC_SHA,TLS_DHE_RSA_WITH_AES_256_CBC_SHA,TLS_DHE_DSS_WITH_AES_256_CBC_SHA,TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256,TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256,TLS_RSA_WITH_AES_128_CBC_SHA256,TLS_ECDH_ECDSA_WITH_AES_128_CBC_SHA256,TLS_ECDH_RSA_WITH_AES_128_CBC_SHA256,TLS_DHE_RSA_WITH_AES_128_CBC_SHA256,TLS_DHE_DSS_WITH_AES_128_CBC_SHA256,TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA,TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA,TLS_RSA_WITH_AES_128_CBC_SHA,TLS_ECDH_ECDSA_WITH_AES_128_CBC_SHA,TLS_ECDH_RSA_WITH_AES_128_CBC_SHA,TLS_DHE_RSA_WITH_AES_128_CBC_SHA,TLS_DHE_DSS_WITH_AES_128_CBC_SHA,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDH_ECDSA_WITH_AES_256_GCM_SHA384,TLS_ECDH_RSA_WITH_AES_256_GCM_SHA384,TLS_DHE_RSA_WITH_AES_256_GCM_SHA384,TLS_DHE_DSS_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDH_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDH_RSA_WITH_AES_128_GCM_SHA256,TLS_DHE_RSA_WITH_AES_128_GCM_SHA256,TLS_DHE_DSS_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_3DES_EDE_CBC_SHA,TLS_ECDHE_RSA_WITH_3DES_EDE_CBC_SHA,SSL_RSA_WITH_3DES_EDE_CBC_SHA,TLS_ECDH_ECDSA_WITH_3DES_EDE_CBC_SHA,TLS_ECDH_RSA_WITH_3DES_EDE_CBC_SHA,SSL_DHE_RSA_WITH_3DES_EDE_CBC_SHA,SSL_DHE_DSS_WITH_3DES_EDE_CBC_SHA,TLS_EMPTY_RENEGOTIATION_INFO_SCSV,TLS_DH_anon_WITH_AES_256_GCM_SHA384,TLS_DH_anon_WITH_AES_128_GCM_SHA256,TLS_DH_anon_WITH_AES_256_CBC_SHA256,TLS_ECDH_anon_WITH_AES_256_CBC_SHA,TLS_DH_anon_WITH_AES_256_CBC_SHA,TLS_DH_anon_WITH_AES_128_CBC_SHA256,TLS_ECDH_anon_WITH_AES_128_CBC_SHA,TLS_DH_anon_WITH_AES_128_CBC_SHA,TLS_ECDH_anon_WITH_3DES_EDE_CBC_SHA,SSL_DH_anon_WITH_3DES_EDE_CBC_SHA,SSL_RSA_WITH_DES_CBC_SHA,SSL_DHE_RSA_WITH_DES_CBC_SHA,SSL_DHE_DSS_WITH_DES_CBC_SHA,SSL_DH_anon_WITH_DES_CBC_SHA,SSL_RSA_EXPORT_WITH_DES40_CBC_SHA,SSL_DHE_RSA_EXPORT_WITH_DES40_CBC_SHA,SSL_DHE_DSS_EXPORT_WITH_DES40_CBC_SHA,SSL_DH_anon_EXPORT_WITH_DES40_CBC_SHA,TLS_RSA_WITH_NULL_SHA256,TLS_ECDHE_ECDSA_WITH_NULL_SHA,TLS_ECDHE_RSA_WITH_NULL_SHA,SSL_RSA_WITH_NULL_SHA,TLS_ECDH_ECDSA_WITH_NULL_SHA,TLS_ECDH_RSA_WITH_NULL_SHA,TLS_ECDH_anon_WITH_NULL_SHA,SSL_RSA_WITH_NULL_MD5,TLS_KRB5_WITH_3DES_EDE_CBC_SHA,TLS_KRB5_WITH_3DES_EDE_CBC_MD5,TLS_KRB5_WITH_DES_CBC_SHA,TLS_KRB5_WITH_DES_CBC_MD5,TLS_KRB5_EXPORT_WITH_DES_CBC_40_SHA,TLS_KRB5_EXPORT_WITH_DES_CBC_40_MD5
`

src/main/java/io/cdap/plugin/http/source/batch/HttpInputFormatProvider.java
aonischuk: awesome idea

src/main/java/io/cdap/plugin/http/source/common/HttpErrorHandler.java
aonischuk: We already have a default in the widget.
`2..:Success,.*:Fail`

src/main/java/io/cdap/plugin/http/source/common/HttpErrorHandlingStrategy.java
aonischuk: My original thought was that something like 500 error code might require retries. While something like 400 will not, since it's an error in client request.

That's why we have those strategies. Do you still think we should remove them?

src/main/java/io/cdap/plugin/http/source/common/HttpErrorHandlingStrategy.java
aonischuk: Means that a certain http code will not be treated as an error. But as fully valid page. Example:
`2..:Success, .*:Fail`

src/main/java/io/cdap/plugin/http/source/batch/HttpBatchSource.java
aonischuk: I just tried to do this, but it makes things a bit to over-complicated when error handling strategy is "SKIP". Since we cannot simply skip in iterator, if has hasNext() already returned true. We can pre-cache of course. I don't really like that. Skipping in transform is super simple on the other hand.

I would prefer to have here "WritableComparable" instead of "Text" when we implement Blob.
Both BytesWritable and Text implement WritableComparable.

What do you think?

src/main/java/io/cdap/plugin/http/source/batch/HttpBatchSource.java
aonischuk: done

src/main/java/io/cdap/plugin/http/source/batch/HttpBatchSource.java
aonischuk: done

src/main/java/io/cdap/plugin/http/source/batch/HttpBatchSource.java
aonischuk: done

src/main/java/io/cdap/plugin/http/source/batch/HttpInputFormatProvider.java
aonischuk: done

src/main/java/io/cdap/plugin/http/source/batch/HttpRecordReader.java
aonischuk: done

src/main/java/io/cdap/plugin/http/source/batch/HttpRecordReader.java
aonischuk: done

src/main/java/io/cdap/plugin/http/source/common/BaseHttpSourceConfig.java
aonischuk: done

src/main/java/io/cdap/plugin/http/source/common/BaseHttpSourceConfig.java
aonischuk: done

src/main/java/io/cdap/plugin/http/source/common/ConfigUtil.java
aonischuk: done

src/main/java/io/cdap/plugin/http/source/common/ConfigUtil.java
aonischuk: done

src/main/java/io/cdap/plugin/http/source/common/ConfigUtil.java
aonischuk: done

src/main/java/io/cdap/plugin/http/source/common/ConfigUtil.java
aonischuk: done

src/main/java/io/cdap/plugin/http/source/common/HttpErrorHandlingStrategy.java
aonischuk: done

src/main/java/io/cdap/plugin/http/source/common/pagination/BaseHttpPaginationIterator.java
aonischuk: done

aonischuk:
@albertshau addressed all the comments. Can you please look?

src/main/java/io/cdap/plugin/http/source/common/HttpErrorHandlingStrategy.java
albertshau: Hm I see, that makes sense. Let's keep it how you have it then.

src/main/java/io/cdap/plugin/http/source/common/HttpErrorHandler.java
albertshau: Yes but somebody can remove those defaults, or the pipeline can be created through the REST API.

src/main/java/io/cdap/plugin/http/source/batch/HttpBatchSource.java
albertshau: I'm not sure I understand why it's hard to skip in the RecordReader, but sending to error seems like it would be impossible. 

Not sure what would be the best way right now, let's revisit when formats are implemented.

src/main/java/io/cdap/plugin/http/source/batch/HttpInputFormatProvider.java
```diff
@@ -17,67 +17,24 @@
...
 import io.cdap.cdap.api.data.batch.InputFormatProvider;
-import io.cdap.plugin.http.source.common.BaseHttpSourceConfig;
 
 import java.util.Map;
 
 /**
  * InputFormatProvider used by cdap to provide configurations to mapreduce job
  */
 public class HttpInputFormatProvider implements InputFormatProvider {
+  public static final String PROPERTY_CONFIG_JSON = "mapreduce.config.json";
```
albertshau: By convention, the mapreduce prefix is for properties from the mapreduce framework/libraries. I think it's better to use something like 'cdap.http.config'

src/main/java/io/cdap/plugin/http/source/common/pagination/PageEntry.java
```diff
@@ -0,0 +1,53 @@
...
+import io.cdap.cdap.etl.api.InvalidEntry;
+
+import java.util.Objects;
+
+/**
+ * A single text entry from a page. Ready to be converted to {@link StructuredRecord} or {@link InvalidEntry}.
+ */
+public class PageEntry {
+  private int httpCode;
+  private String body;
```
albertshau: when we introduce formats, we'll have to change this but we'll leave it for that PR.

src/main/java/io/cdap/plugin/http/source/common/pagination/PageEntry.java
```diff
@@ -0,0 +1,53 @@
...
+import io.cdap.cdap.api.data.format.StructuredRecord;
+import io.cdap.cdap.etl.api.InvalidEntry;
+
+import java.util.Objects;
+
+/**
+ * A single text entry from a page. Ready to be converted to {@link StructuredRecord} or {@link InvalidEntry}.
+ */
+public class PageEntry {
+  private int httpCode;
```
albertshau: these should be final.

albertshau (commented):
small comments, otherwise lgtm. Please fix and squash, then i'll merge.

src/main/java/io/cdap/plugin/http/source/common/pagination/PageEntry.java
aonischuk: done

src/main/java/io/cdap/plugin/http/source/batch/HttpInputFormatProvider.java
aonischuk: done

src/main/java/io/cdap/plugin/http/source/common/HttpErrorHandler.java
aonischuk: Done

PR #11, opened 2019-07-18
CDAP-15524. HTTP Batch Source. Pagination. Http client implementation

aonischuk:
1. Implemented all pagination types. Including custom pagination (via python code)
2. Implemented http client wrapper. Which includes proxy, basic auth, oauth2 request, TSL/SSL functionality, timeout handling.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -75,17 +75,22 @@
   </repositories>
 
   <properties>
+    <awaitility.version>3.1.6</awaitility.version>
     <cdap.version>6.0.0</cdap.version>
-    <hadoop.version>2.3.0</hadoop.version>
-    <junit.version>4.11</junit.version>
-    <hydrator.version>2.2.0</hydrator.version>
-    <widgets.dir>widgets</widgets.dir>
+    <commons.version>3.9</commons.version>
+    <common.codec.version>1.12</common.codec.version>
     <docs.dir>docs</docs.dir>
-    <app.parents>
-      system:cdap-data-pipeline[6.0.0-SNAPSHOT,7.0.0-SNAPSHOT),system:cdap-data-streams[6.0.0-SNAPSHOT,7.0.0-SNAPSHOT)
-    </app.parents>
+    <gson.version>2.8.5</gson.version>
+    <hadoop.version>2.3.0</hadoop.version>
+    <httpcomponents.version>4.5.9</httpcomponents.version>
+    <hydrator.version>2.2.0</hydrator.version>
+    <jackson.version>2.9.9</jackson.version>
+    <junit.version>4.11</junit.version>
+    <jython.version>2.7.1</jython.version>
     <!-- this is here because project.basedir evaluates to null in the script build step -->
     <main.basedir>${project.basedir}</main.basedir>
+    <mockito.version>1.10.19</mockito.version>
+    <unxml.version>0.9</unxml.version>
   </properties>
 
   <dependencies>
@@ -103,31 +108,23 @@
     </dependency>
     <dependency>
       <groupId>io.cdap.cdap</groupId>
-      <artifactId>cdap-unit-test</artifactId>
+      <artifactId>cdap-formats</artifactId>
       <version>${cdap.version}</version>
-      <scope>test</scope>
       <exclusions>
         <exclusion>
-          <groupId>asm</groupId>
-          <artifactId>asm</artifactId>
+          <groupId>org.apache.avro</groupId>
+          <artifactId>avro</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>io.thekraken</groupId>
+          <artifactId>grok</artifactId>
         </exclusion>
       </exclusions>
     </dependency>
     <dependency>
-      <groupId>com.google.guava</groupId>
-      <artifactId>guava</artifactId>
-      <version>13.0.1</version>
-    </dependency>
-    <dependency>
-      <groupId>io.cdap.cdap</groupId>
-      <artifactId>cdap-data-pipeline</artifactId>
-      <version>${cdap.version}</version>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>io.cdap.cdap</groupId>
-      <artifactId>cdap-formats</artifactId>
-      <version>${cdap.version}</version>
+      <groupId>io.cdap.plugin</groupId>
+      <artifactId>hydrator-common</artifactId>
+      <version>${hydrator.version}</version>
     </dependency>
     <dependency>
       <groupId>org.apache.hadoop</groupId>
@@ -241,16 +238,84 @@
         </exclusion>
       </exclusions>
     </dependency>
+
     <dependency>
-      <groupId>io.cdap.plugin</groupId>
-      <artifactId>hydrator-common</artifactId>
-      <version>${hydrator.version}</version>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-lang3</artifactId>
+      <version>${commons.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>commons-codec</groupId>
+      <artifactId>commons-codec</artifactId>
+      <version>${common.codec.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.awaitility</groupId>
+      <artifactId>awaitility</artifactId>
+      <version>${awaitility.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>httpclient</artifactId>
+      <version>${httpcomponents.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.code.gson</groupId>
+      <artifactId>gson</artifactId>
+      <version>${gson.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+      <version>${jackson.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.dataformat</groupId>
+      <artifactId>jackson-dataformat-xml</artifactId>
+      <version>${jackson.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>com.nerdforge</groupId>
+      <artifactId>unxml</artifactId>
+      <version>${unxml.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.python</groupId>
+      <artifactId>jython-standalone</artifactId>
+      <version>${jython.version}</version>
+    </dependency>
+
+
+    <!-- tests -->
+    <dependency>
+      <groupId>io.cdap.cdap</groupId>
+      <artifactId>cdap-etl-proto</artifactId>
+      <version>${cdap.version}</version>
+      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>io.cdap.cdap</groupId>
       <artifactId>hydrator-test</artifactId>
       <version>${cdap.version}</version>
-      <scope>provided</scope>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>io.cdap.cdap</groupId>
+      <artifactId>cdap-data-pipeline</artifactId>
+      <version>${cdap.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>${junit.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-all</artifactId>
+      <version>${mockito.version}</version>
+      <scope>test</scope>
     </dependency>
   </dependencies>
 
--- a/src/main/java/io/cdap/plugin/http/source/common/http/HttpClient.java
+++ b/src/main/java/io/cdap/plugin/http/source/common/http/HttpClient.java
@@ -1,0 +1,166 @@
+/*
+ * Copyright © 2019 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.cdap.plugin.http.source.common.http;
+
+import com.google.common.base.Charsets;
+import com.google.common.base.Strings;
+import io.cdap.plugin.http.source.common.BaseHttpSourceConfig;
+import org.apache.http.Header;
+import org.apache.http.HttpHost;
+import org.apache.http.auth.AuthScope;
+import org.apache.http.auth.UsernamePasswordCredentials;
+import org.apache.http.client.CredentialsProvider;
+import org.apache.http.client.config.RequestConfig;
+import org.apache.http.client.methods.CloseableHttpResponse;
+import org.apache.http.client.methods.HttpEntityEnclosingRequestBase;
+import org.apache.http.entity.StringEntity;
+import org.apache.http.impl.client.BasicCredentialsProvider;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.impl.client.HttpClientBuilder;
+import org.apache.http.impl.client.HttpClients;
+import org.apache.http.message.BasicHeader;
+
+import java.io.Closeable;
+import java.io.IOException;
+import java.net.URI;
+import java.util.ArrayList;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * An http client used to get data from given url. It follows the configurations from {@link BaseHttpSourceConfig}
+ */
+public class HttpClient implements Closeable {
+  private final Map<String, String> headers;
+  private final BaseHttpSourceConfig config;
+  private final StringEntity requestBody;
+  private CloseableHttpClient httpClient;
+
+  public HttpClient(BaseHttpSourceConfig config) {
+    this.config = config;
+    this.headers = config.getHeadersMap();
+
+    String requestBodyString = config.getRequestBody();
+    if (requestBodyString != null) {
+      requestBody = new StringEntity(requestBodyString, Charsets.UTF_8.toString());
+    } else {
+      requestBody = null;
+    }
+  }
+
+  /**
+   * Executes HTTP request with parameters configured in plugin config and returns response.
+   * Is called to load every page by pagination iterator.
+   *
+   * @param uri URI of resource
+   * @return a response object
+   * @throws IOException in case of a problem or the connection was aborted
+   */
+  public CloseableHttpResponse executeHTTP(String uri) throws IOException {
+    // lazy init. So we are able to initialize the class for different checks during validations etc.
+    if (httpClient == null) {
+      httpClient = createHttpClient();
+    }
+
+    HttpEntityEnclosingRequestBase request = new HttpRequest(URI.create(uri), config.getHttpMethod());
+
+    if (requestBody != null) {
+      request.setEntity(requestBody);
+    }
+
+    return httpClient.execute(request);
+  }
+
+  @Override
+  public void close() throws IOException {
+    if (httpClient != null) {
+      httpClient.close();
+    }
+  }
+
+  private CloseableHttpClient createHttpClient() throws IOException {
+    HttpClientBuilder httpClientBuilder = HttpClientBuilder.create();
+    httpClientBuilder.setSSLSocketFactory(new SSLConnectionSocketFactoryCreator(config).create());
+
+    // set timeouts
+    Long connectTimeoutMillis = TimeUnit.SECONDS.toMillis(config.getConnectTimeout());
+    Long readTimeoutMillis = TimeUnit.SECONDS.toMillis(config.getReadTimeout());
+    RequestConfig.Builder requestBuilder = RequestConfig.custom();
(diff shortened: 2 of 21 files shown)


albertshau:
should use mocks for the unit tests, otherwise they can break due to external changes. See https://github.com/cdapio/hydrator-plugins/blob/develop/spark-plugins/src/test/java/io/cdap/plugin/spark/test/SparkPluginTest.java#L119 for an example.

pom.xml
```diff
@@ -75,17 +75,22 @@
...
-    <app.parents>
-      system:cdap-data-pipeline[6.0.0-SNAPSHOT,7.0.0-SNAPSHOT),system:cdap-data-streams[6.0.0-SNAPSHOT,7.0.0-SNAPSHOT)
-    </app.parents>
+    <jackson.version>2.9.9</jackson.version>
+    <junit.version>4.11</junit.version>
+    <jython.version>2.7.1</jython.version>
     <!-- this is here because project.basedir evaluates to null in the script build step -->
     <main.basedir>${project.basedir}</main.basedir>
+    <unxml.version>0.9</unxml.version>
+    <widgets.dir>widgets</widgets.dir>
```
albertshau: this can be removed. It's a relic from when the build was using javascript instead of the cdap maven plugin

pom.xml
```diff
@@ -103,31 +108,13 @@
...
-    </dependency>
-    <dependency>
-      <groupId>com.google.guava</groupId>
-      <artifactId>guava</artifactId>
-      <version>13.0.1</version>
-    </dependency>
-    <dependency>
-      <groupId>io.cdap.cdap</groupId>
-      <artifactId>cdap-data-pipeline</artifactId>
+      <artifactId>cdap-formats</artifactId>
```
albertshau, line 111: I think this pulls in stuff like avro and grok. Since we're not using those, let's add exclusions for them, which will make the jar a little bit smaller.

pom.xml
```diff
@@ -241,16 +228,100 @@
...
+      <groupId>org.python</groupId>
+      <artifactId>jython-standalone</artifactId>
+      <version>${jython.version}</version>
+    </dependency>
+
+
+    <!-- tests -->
+    <dependency>
+      <groupId>io.cdap.cdap</groupId>
+      <artifactId>cdap-metadata-spi</artifactId>
```
albertshau: why is this needed?

pom.xml
```diff
@@ -241,16 +228,100 @@
...
+      <version>${cdap.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>io.cdap.cdap</groupId>
+      <artifactId>cdap-common</artifactId>
```
albertshau: why is this needed?

src/main/java/io/cdap/plugin/http/source/common/http/HttpClient.java
```diff
@@ -0,0 +1,167 @@
...
+
+  private CloseableHttpClient createHttpClient(BaseHttpSourceConfig config) throws IOException, URISyntaxException {
+    HttpClientBuilder httpClientBuilder = HttpClientBuilder.create();
+    httpClientBuilder.setSSLSocketFactory(new SSLConnectionSocketFactoryCreator(config).create());
+
+    // set timeouts
+    Long connectTimeoutMillis = TimeUnit.SECONDS.toMillis(config.getConnectTimeout());
+    Long readTimeoutMillis = TimeUnit.SECONDS.toMillis(config.getReadTimeout());
+    RequestConfig.Builder requestBuilder = RequestConfig.custom();
+    requestBuilder.setSocketTimeout(readTimeoutMillis.intValue());
```
albertshau, line 102: if these should be ints, config.getConnectTimeout() should also return an int. Otherwise there can be weird behavior due to overflow.

src/main/java/io/cdap/plugin/http/source/common/http/HttpClient.java
```diff
@@ -0,0 +1,167 @@
...
+
+    String requestBodyString = config.getRequestBody();
+    if (requestBodyString != null) {
+      requestBody = new StringEntity(requestBodyString, Charsets.UTF_8.toString());
+    } else {
+      requestBody = null;
+    }
+  }
+
+  public CloseableHttpResponse executeHTTP(String uri) throws IOException {
```
albertshau, line 72: is this meant to be called just once or can it be called more than once? Please add a javadoc to describe expected usage.

src/main/java/io/cdap/plugin/http/source/common/http/HttpClient.java
```diff
@@ -0,0 +1,167 @@
...
+  public void close() throws IOException {
+    if (httpClient != null) {
+      httpClient.close();
+    }
+  }
+
+  private void initialize() {
+    try {
+      httpClient = createHttpClient(config);
+    } catch (IOException | URISyntaxException e) {
```
albertshau: these are two pretty different types of errors, it's better not to wrap them in a runtime exception, but to let them propagate up.

I'm also confused about why createHttpClient() throws a URISyntaxException. Which method call is throwing it? Ideally this isn't an exception we would have to deal with. It would be better if this class gets passed in URI or URL objects instead of dealing with strings, with validation of those done outside this class. 

In general, validation of user input should happen as early as possible, with classes later in the logic using classes that are guaranteed to be valid.

src/main/java/io/cdap/plugin/http/source/common/http/SSLConnectionSocketFactoryCreator.java
```diff
@@ -0,0 +1,130 @@
...
+    KeyStoreException, IOException, UnrecoverableKeyException {
+
+    KeyStore keystore = loadKeystore(config.getKeystoreFile(), config.getKeystoreType().name(),
+                                     config.getKeystorePassword());
+
+    String keyStorePassword = config.getKeystorePassword();
+
+    // we have to manually fall back to default keystore. SSLContext won't provide such a functionality.
+    if (keystore == null) {
+      String keyStore = System.getProperty("javax.net.ssl.keyStore");
```
albertshau, line 80: How do these properties get set?

src/main/java/io/cdap/plugin/http/source/common/pagination/BaseHttpPaginationIterator.java
```diff
@@ -15,15 +15,182 @@
...
+   * @return true if page still has elements to iterate. Otherwise it will load next page.
+   * False if no more pages to load or the page loaded has no elements.
+   */
+  protected boolean ensurePageIterable() {
+    try {
+      if (page == null || !page.hasNext()) {
+        page = getNextPage();
+      }
+
+      return (page != null && page.hasNext());
```
albertshau: nit: can remove the outer parentheses

src/main/java/io/cdap/plugin/http/source/common/pagination/BaseHttpPaginationIterator.java
```diff
@@ -15,15 +15,182 @@
...
+    }
+
+    try {
+      Awaitility
+        .await().with()
+        .pollInterval(pollInterval)
+        .timeout(config.getMaxRetryDuration(), TimeUnit.SECONDS)
+        .until(this::visitPageAndCheckStatusCode);
+    } catch (ConditionTimeoutException ex) {
+      // Retries failed. We don't need to do anything here. This will ba handled using httpStatusCode below.
```
albertshau: typo: ba -> be

src/main/java/io/cdap/plugin/http/source/common/pagination/BaseHttpPaginationIterator.java
albertshau: this will return a false negative if it's possible to have an empty Page. To protect against this, it's better to loop above until we get a page with entries.

src/main/java/io/cdap/plugin/http/source/common/pagination/BaseHttpPaginationIterator.java
```diff
@@ -15,15 +15,182 @@
...
+    }
+
+    response = httpClient.executeHTTP(nextPageUrl);
+    httpStatusCode = response.getStatusLine().getStatusCode();
+    HttpErrorHandlingStrategy errorHandlingStrategy = httpErrorHandler.getErrorHandlingStrategy(httpStatusCode);
+
+    return !errorHandlingStrategy.shouldRetry();
+  }
+
+  protected BasePage getNextPage() throws IOException, InterruptedException {
```
albertshau: annotate as nullabe

src/main/java/io/cdap/plugin/http/source/common/pagination/BaseHttpPaginationIterator.java
```diff
@@ -15,15 +15,182 @@
...
+    }
+
+    // response being null, means it's the first page we are loading
+    if (response != null && config.getWaitTimeBetweenPages() != null) {
+      Thread.sleep(config.getWaitTimeBetweenPages());
+    }
+
+    LOG.debug("Fetching '{}'", nextPageUrl);
+
+    PollInterval pollInterval;
```
albertshau: it looks like this could be a private final that is set once in the constructor.

src/main/java/io/cdap/plugin/http/source/common/pagination/BaseHttpPaginationIterator.java
```diff
@@ -15,15 +15,182 @@
...
+        .await().with()
+        .pollInterval(pollInterval)
+        .timeout(config.getMaxRetryDuration(), TimeUnit.SECONDS)
+        .until(this::visitPageAndCheckStatusCode);
+    } catch (ConditionTimeoutException ex) {
+      // Retries failed. We don't need to do anything here. This will ba handled using httpStatusCode below.
+    }
+
+    HttpEntity reponseEntity = response.getEntity();
+    byte[] responseBytes = EntityUtils.toByteArray(reponseEntity);
```
albertshau, line 114: we should not be loading the entire contents into memory unless it's an error. Under normal circumstances the contents should be read as an InputStream.

src/main/java/io/cdap/plugin/http/source/common/pagination/BaseHttpPaginationIterator.java
```diff
@@ -15,15 +15,182 @@
...
+    if (config.getFormat().equals(PageFormat.BLOB)) {
```
albertshau: Format should only apply on success. Most of the time, an error response is not formatted the same way a successful response is formatted.

src/main/java/io/cdap/plugin/http/source/common/pagination/JythonPythonExecutor.java
```diff
@@ -0,0 +1,96 @@
...
+
+    try {
+      interpreter.exec(script);
+    } catch (PySyntaxError e) {
+      if (Strings.isNullOrEmpty(e.getMessage())) {
+        throw new RuntimeException(
+          "Unknown syntax error occurred while interpreting python code. Please make sure syntax is correct.", e);
+      }
+      throw new RuntimeException(
+        String.format("Syntax error occurred while interpreting python code: '%s'" +
```
albertshau: Should not surround the exception message with text, as it makes it hard to distinguish between what comes from the exception and what comes from this line. Better to just have:

Syntax error occurred while interpreting python code: '%s'

src/main/java/io/cdap/plugin/http/source/common/pagination/LinkInResponseBodyPagionationIterator.java
```diff
@@ -0,0 +1,66 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+
+/**
+ * Every page contains a next page url. This pagination type is only supported for JSON and XML formats.
+ * Pagination happens until no next page field is present or until page contains no elements.
+ */
+public class LinkInResponseBodyPagionationIterator extends BaseHttpPaginationIterator {
```
albertshau: typo: Pagio -> Pagi

src/main/java/io/cdap/plugin/http/source/common/pagination/NonePagionationIterator.java
```diff
@@ -0,0 +1,43 @@
...
+import io.cdap.plugin.http.source.common.BaseHttpSourceConfig;
+import io.cdap.plugin.http.source.common.pagination.page.BasePage;
+import org.apache.http.client.methods.CloseableHttpResponse;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Returns a single page of data
+ */
+public class NonePagionationIterator extends BaseHttpPaginationIterator {
```
albertshau: typo: Pagio -> Pagi

albertshau (commented):
Can you move the format stuff to a separate PR? 2000 lines is a lot and there aren't any unit tests yet. Pagination and format shouldn't share much code.

albertshau:
After taking a look, you also shouldn't need to setup an actual http server for the unit tests. The pagination classes can all be tested using a mock HttpClient that returns a response that you set in the test.

src/main/java/io/cdap/plugin/http/source/common/http/HttpClient.java
aonischuk: URISyntaxException is thown by OAuthUtil.getAccessTokenByRefreshToken() Which forms URI for retrieving accessToken by RefreshToken.

src/main/java/io/cdap/plugin/http/source/common/http/SSLConnectionSocketFactoryCreator.java
aonischuk: They are unset by default. But to change default keyStore user can set them via "java -D....".

This code is there just not to break default ssl settings behavior in java when javax.net.ssl.* properties are set.

src/main/java/io/cdap/plugin/http/source/common/pagination/BaseHttpPaginationIterator.java
aonischuk: I don't think Page instance can  be null. See PageFactory#createInstance it cannot return null. Null can only be returned from getNextPage() as a sign that there are no more pages to load.

src/main/java/io/cdap/plugin/http/source/common/pagination/BaseHttpPaginationIterator.java
aonischuk: At some point we will need to read whole byte array into memory. Since it needs to be written to structured record "body" field (field type is "bytes"). Am I missing something?

src/main/java/io/cdap/plugin/http/source/common/http/HttpClient.java
aonischuk: done

src/main/java/io/cdap/plugin/http/source/common/pagination/BaseHttpPaginationIterator.java
aonischuk: done

pom.xml
aonischuk: done

pom.xml
aonischuk: done

pom.xml
aonischuk: done

pom.xml
aonischuk: done

src/main/java/io/cdap/plugin/http/source/common/http/HttpClient.java
aonischuk: done. However note that TimeUnit.SECONDS.toMillis still returns long. So conversion to int remains.

src/main/java/io/cdap/plugin/http/source/common/http/HttpClient.java
aonischuk: done

src/main/java/io/cdap/plugin/http/source/common/pagination/BaseHttpPaginationIterator.java
aonischuk: done (removed the outer parentheses)

src/main/java/io/cdap/plugin/http/source/common/pagination/BaseHttpPaginationIterator.java
aonischuk: done

src/main/java/io/cdap/plugin/http/source/common/pagination/BaseHttpPaginationIterator.java
aonischuk: done

src/main/java/io/cdap/plugin/http/source/common/pagination/BaseHttpPaginationIterator.java
aonischuk: done

src/main/java/io/cdap/plugin/http/source/common/pagination/JythonPythonExecutor.java
aonischuk: done

src/main/java/io/cdap/plugin/http/source/common/pagination/NonePagionationIterator.java
aonischuk: done

src/main/java/io/cdap/plugin/http/source/common/pagination/LinkInResponseBodyPagionationIterator.java
aonischuk: done

aonischuk:
1. Removed the format stuff from PR.
2. Regarding UT: I have added unit tests
3. Thank you for reviewing so much code lately.

@albertshau I have addressed your comments. Can we please merge if everything looks good?

src/main/java/io/cdap/plugin/http/source/common/pagination/BaseHttpPaginationIterator.java
albertshau: That's only for the 'blob' format. Everything else should not. You can ignore this now for now, since format has been moved out.

src/main/java/io/cdap/plugin/http/source/common/pagination/BaseHttpPaginationIterator.java
albertshau: Empty page meaning one where page.hasNext() is false.

If page.hasNext() is false and getNextPage() would have returned a non-null, then this will terminate early.

albertshau:
One remaining comment, please fix and squash, then i'll merge.

aonischuk:
@albertshau squashed the commits. Please feel free to merge.

Ps. also added pagination unit tests.

src/main/java/io/cdap/plugin/http/source/common/pagination/BaseHttpPaginationIterator.java
```diff
@@ -15,15 +15,194 @@
...
+   * @return true if page still has elements to iterate. Otherwise it will load next page.
+   * False if no more pages to load or the page loaded has no elements.
+   */
+  protected boolean ensurePageIterable() {
+    try {
+      if (page == null || !page.hasNext()) {
+        page = getNextPage();
+      }
+
+      return page != null && page.hasNext();
```
albertshau, line 165: comment here is still not addressed.

src/test/java/io/cdap/plugin/http/source/common/pagination/PaginationIteratorTest.java
```diff
@@ -0,0 +1,288 @@
+package io.cdap.plugin.http.source.common.pagination;
```
albertshau, line 17: missing copyright

src/main/java/io/cdap/plugin/http/source/common/pagination/BaseHttpPaginationIterator.java
aonischuk: Empty page (with no elements) in pagination APIs usually means that you should stop paginating otherwise you'll get into endless loop.

Example API:
https://services.odata.org/V3/Northwind/Northwind.svc/Customers?$inlinecount=allpages&$top=5&$skip=80

You can go to page 100 200 500 1000 which are empty and it's endless. All with no elements.
Page 10.000:
https://services.odata.org/V3/Northwind/Northwind.svc/Customers?$inlinecount=allpages&$top=5&$skip=10000

aonischuk:
1. Answered your remaining comment
2. Added pagination unit tests.
3. Squashed the commit.

@albertshau please feel free to merge.

albertshau:
Did you leave an answer about my comment? Didn't see it.

src/main/java/io/cdap/plugin/http/source/common/pagination/BaseHttpPaginationIterator.java
aonischuk: Empty page (with no elements) in pagination APIs usually means that you should stop paginating otherwise you'll get into endless loop.

Example API:
https://services.odata.org/V3/Northwind/Northwind.svc/Customers?$inlinecount=allpages&$top=5&$skip=80

You can go to page 100 200 500 1000 which are empty and it's endless. All with no elements.
Page 10.000:
https://services.odata.org/V3/Northwind/Northwind.svc/Customers?$inlinecount=allpages&$top=5&$skip=10000

aonischuk: that's why I don't think we need any change here. Please correct me if I am wrong.

src/test/java/io/cdap/plugin/http/source/common/pagination/PaginationIteratorTest.java
aonischuk: done

aonischuk:
1. Added copyright
2. Answered the comment regarding empty page getNext().

@albertshau please re-check/merge

src/main/java/io/cdap/plugin/http/source/common/pagination/BaseHttpPaginationIterator.java
albertshau: I see, makes sense

aonischuk:
Thanks!

PR #12, opened 2019-07-19
CDAP-15524. HTTP Batch Source. Formats Implementation.

aonischuk:
Implement all the formats. Which includes: JSON, XML, TSV, CSV, TEXT, BLOB.


src/main/java/io/cdap/plugin/http/source/common/pagination/page/PageFactory.java
```diff
@@ -25,6 +25,24 @@
 public class PageFactory {
   public static BasePage createInstance(BaseHttpSourceConfig config, String responseBody,
                                         HttpErrorHandlingStrategy postRetryStrategy) {
-    throw new IllegalStateException("Not yet implemented"); // TODO: implement this
+    // if response is erroneous
+    if (!postRetryStrategy.equals(HttpErrorHandlingStrategy.SUCCESS)) {
```
albertshau: this seems like it should be done by the caller and not in this method.

src/main/java/io/cdap/plugin/http/source/common/pagination/page/PageFactory.java
```diff
@@ -25,6 +25,24 @@
...
+      case JSON:
+        return new JsonPage(config, responseBody);
+      case XML:
+        return new XmlPage(config, responseBody);
+      case TSV:
+      case CSV:
+      case TEXT:
+        return new RecordPerLinePage(config, responseBody);
+      case BLOB:
+        return new WholeTextPage(config, responseBody);
```
albertshau: blob in the file source means the output schema is bytes, not a string. I think we should make this behavior similarly. This is useful if the API is returning sound clips or images or something like that.

If the blob is actually a string, it can be converted to one in a later transform.

src/main/java/io/cdap/plugin/http/source/common/pagination/page/PageFactory.java
```diff
@@ -25,6 +25,24 @@
...
+
+    switch(config.getFormat()) {
+      case JSON:
+        return new JsonPage(config, responseBody);
+      case XML:
+        return new XmlPage(config, responseBody);
+      case TSV:
+      case CSV:
+      case TEXT:
+        return new RecordPerLinePage(config, responseBody);
```
albertshau: These formats do not require loading the entire page in memory. Instead of taking the response body as a string, this method should take whatever object you can get the InputStream from. The Page objects should also be Closeable so that they can close their InputStreams when finished.

Conceptually, json and xml should not require it either, though I don't know if the libraries we're using allow us to do that easily.

src/main/java/io/cdap/plugin/http/source/common/record/LinesToRecordConverter.java
```diff
@@ -0,0 +1,67 @@
...
+public class LinesToRecordConverter implements BaseStringToRecordConverter {
+  private final Schema schema;
+  private final RecordFormat<ByteBuffer, StructuredRecord> recordFormat;
+
+  public LinesToRecordConverter(BaseHttpSourceConfig config, Schema schema) {
+    this.schema = schema;
+
+    FormatSpecification spec = new FormatSpecification(config.getFormat().getValue(), schema, new HashMap<>());
+    try {
+      this.recordFormat = RecordFormats.createInitializedFormat(spec);
```
albertshau: RecordFormats is an old thing, originally written for applying formats to CDAP Streams (which were removed in 6.0.0). This is overkill for changing a line into a StructuredRecord for just the 'text', 'csv' and 'tsv' formats. You can use StructuredRecordStringConverter.fromDelimitedString instead.

src/main/java/io/cdap/plugin/http/source/common/pagination/page/JsonPage.java
```diff
@@ -0,0 +1,83 @@
...
+    fieldsMapping = config.getFullFieldsMapping();
+  }
+
+  @Override
+  public boolean hasNext() {
+    return iterator.hasNext();
+  }
+
+  @Override
+  public String next() {
```
albertshau: Instead of making these pages return Strings and having a separate record converter class that converts Strings to StructuredRecords, it seems better to have Page return StructuredRecord.

It doesn't make sense to convert to String in some situations, like for the blob format where the response is returned as bytes. Here it would also be possible to convert the JsonObject directly into a StructuredRecord rather than to a string first.

src/main/java/io/cdap/plugin/http/source/common/pagination/page/JsonPage.java
```diff
@@ -0,0 +1,83 @@
...
+    JsonObject currentJsonObject = iterator.next().getAsJsonObject();
+
+    JsonObject resultJson = new JsonObject();
+    for (Map.Entry<String, String> entry : fieldsMapping.entrySet()) {
+      String schemaFieldName = entry.getKey();
+      String fieldPath = insideElementJsonPathPart + "/" + StringUtils.stripStart(entry.getValue(), "/");
```
albertshau, line 136: what is happening here? Can you add a comment with an example?

src/main/java/io/cdap/plugin/http/source/common/pagination/page/JsonPage.java
```diff
@@ -0,0 +1,83 @@
...
+   * Get primitive from json by json path or return null if not found.
+   * If element found is not a primitive (object or array) exception is thrown.
+   *
+   * @param path a json path. E.g. "/city/schools/students"
+   * @return a primitive converted to string
+   */
+  @Override
+  public String getPrimitiveByPath(String path) {
+    JSONUtil.JsonQueryResponse queryResponse = JSONUtil.getJsonElementByPath(
+      json, path);
```
albertshau: nit: looks like it could on previous line

src/main/java/io/cdap/plugin/http/source/common/pagination/page/JsonPage.java
```diff
@@ -0,0 +1,83 @@
...
+
+  /**
+   * Get primitive from json by json path or return null if not found.
+   * If element found is not a primitive (object or array) exception is thrown.
+   *
+   * @param path a json path. E.g. "/city/schools/students"
+   * @return a primitive converted to string
+   */
+  @Override
+  public String getPrimitiveByPath(String path) {
```
albertshau, line 161: This is a BasePage method? What does it mean for the delimited formats?

I also don't see where this is used, when is it called?

src/main/java/io/cdap/plugin/http/source/common/pagination/page/JSONUtil.java
```diff
@@ -28,5 +34,151 @@ public static JsonObject toJsonObject(String text) {
...
+      new String[0],
+      currentElement
+    );
+  }
+
+  /**
+   * A class which contains information regarding results of searching an element in json by json path.
+   */
+  public static class JsonQueryResponse {
+    private final String[] retrivedPathParts;
```
albertshau: typos: retrieved, unretrieved

src/main/java/io/cdap/plugin/http/source/common/pagination/page/JSONUtil.java
```diff
@@ -28,5 +34,151 @@ public static JsonObject toJsonObject(String text) {
...
+    public boolean isFullyRetrieved() {
+      return (unretrivedPathParts.length == 0);
+    }
+
+    /**
+     * Fails if json path was not fully successfully retrieved till the last element.
+     */
+    public void assertFullyRetrieved() {
+      if (!isFullyRetrieved()) {
+        throw new IllegalStateException(String.format(
```
albertshau: this isn't really an illegal state, IllegalArgument fits a little better

src/main/java/io/cdap/plugin/http/source/common/pagination/page/JSONUtil.java
```diff
@@ -28,5 +34,151 @@ public static JsonObject toJsonObject(String text) {
...
+
+  /**
+   * Find an element by jsonPath in given json object. If element not found, information about the search is returned.
+   * Like until which element json path evaluation was successful.
+   *
+   * @param jsonObject a jsonObject on which the search should be performed
+   * @param jsonPath a slash separated path. E.g. "/bookstore/books"
+   * @return an object containing information about search results, success/failure.
+   */
+  public static JsonQueryResponse getJsonElementByPath(JsonObject jsonObject, String jsonPath) {
```
albertshau, line 49: can you add unit tests for this method

src/main/java/io/cdap/plugin/http/source/common/pagination/page/JsonPage.java
```diff
@@ -0,0 +1,83 @@
...
+  private final String insideElementJsonPathPart;
+  private final Iterator<JsonElement> iterator;
+  private final JsonObject json;
+  private final Map<String, String> fieldsMapping;
+
+  JsonPage(BaseHttpSourceConfig config, String responseBody) {
+    json = JSONUtil.toJsonObject(responseBody);
+    JSONUtil.JsonQueryResponse queryResponse = JSONUtil.getJsonElementByPath(json, config.getResultPath());
+    insideElementJsonPathPart = queryResponse.getUnretrivedPath();
+    iterator = queryResponse.getAsJsonArray().iterator();
```
albertshau: We should support the case that the path points to either an object or an array.

I imagine there could be situations where the structure is pretty complex so they want to just get a big object and process it more specially in a transform later in the pipeline.

src/main/java/io/cdap/plugin/http/source/common/pagination/page/JsonPage.java
aonischuk: The problem is emitting and skipping errors in this case.

src/main/java/io/cdap/plugin/http/source/common/pagination/page/PageFactory.java
aonischuk: It is converted to "bytes" already. See BlobToRecordConverter#getRecord

src/main/java/io/cdap/plugin/http/source/common/pagination/page/JsonPage.java
albertshau: we might need to change the InputFormat so that it returns Page. That way you get the Page in the transform method, where you have the ability to emit errors.

src/main/java/io/cdap/plugin/http/source/common/pagination/page/PageFactory.java
albertshau: I see, got confused by the fact that Pages returned strings, which then get converted later.

src/main/java/io/cdap/plugin/http/source/common/pagination/page/JsonPage.java
aonischuk: This is called by pagination iterators: "Link In Response Body" and "Link in Response Token". To retrieve next page url/token. 

Delimited and text types will fail with `Page format '%s' does not support searching by path` when this is called.

However this exception will never actually happen.
Since if user chooses pagination via "Link In Response Body" or "Link in Response Token" and type like text,csv,tsv,blob config validation will fail early.

src/main/java/io/cdap/plugin/http/source/common/pagination/page/PageFactory.java
aonischuk: I have changed this to if(isError) hopefully that is what you meant.

src/main/java/io/cdap/plugin/http/source/common/pagination/page/PageFactory.java
aonischuk: Done

src/main/java/io/cdap/plugin/http/source/common/pagination/page/PageFactory.java
aonischuk: Done for text , csv, txt. 

For Json and Xml unfortunately we have to firstly convert all the string to a jsonObject or xml node. Which requires everything to be parsed.

src/main/java/io/cdap/plugin/http/source/common/record/LinesToRecordConverter.java
aonischuk: Done

src/main/java/io/cdap/plugin/http/source/common/pagination/page/JsonPage.java
aonischuk: Done. However note: it required a lot of extra code to make http error handling and exceptions error handling possible

src/main/java/io/cdap/plugin/http/source/common/pagination/page/JsonPage.java
aonischuk: Added a javadoc to this method with an example

src/main/java/io/cdap/plugin/http/source/common/pagination/page/JsonPage.java
aonischuk: done

src/main/java/io/cdap/plugin/http/source/common/pagination/page/JSONUtil.java
aonischuk: Done

src/main/java/io/cdap/plugin/http/source/common/pagination/page/JSONUtil.java
aonischuk: done

src/main/java/io/cdap/plugin/http/source/common/pagination/page/JSONUtil.java
aonischuk: done

src/main/java/io/cdap/plugin/http/source/common/pagination/page/JsonPage.java
aonischuk: Done

aonischuk:
@albertshau I have addressed all the comments. Can you please merge if everything looks fine?

src/main/java/io/cdap/plugin/http/source/common/pagination/entry/PageEntry.java
```diff
@@ -0,0 +1,61 @@
...
+import java.io.Serializable;
+import javax.annotation.Nullable;
+
+/**
+ * Represents a single record a page with an additional information.
+ * Or an error entry.
+ *
+ * This is returned as a final result of pages fetching.
+ */
+public abstract class PageEntry implements Serializable {
```
albertshau: why Serializable?

src/main/java/io/cdap/plugin/http/source/common/pagination/entry/PageEntry.java
```diff
@@ -0,0 +1,61 @@
...
+  protected static final String ERROR_SCHEMA_BODY_PROPERTY = "body";
+  private static final Schema ERROR_SCHEMA = Schema.recordOf("error",
+                                                             Schema.Field.of(ERROR_SCHEMA_BODY_PROPERTY,
+                                                                             Schema.of(Schema.Type.STRING))
+  );
+
+  protected final String stringVersionOfRecord;
+
+  public PageEntry(String stringVersionOfRecord) {
```
albertshau: There should not be a string representation of a record anywhere in a base class. The fact that a record can come from a string is something specific to different formats.

src/main/java/io/cdap/plugin/http/source/common/pagination/BaseHttpPaginationIterator.java
```diff
@@ -185,7 +174,11 @@ public PageEntry next() {
       throw new NoSuchElementException("No more pages to load.");
     }
 
-    return new PageEntry(httpStatusCode, page.next());
+    try {
+      return new PageSuccessEntry(httpStatusCode, page.next(), page.getStringVersionOfRecord());
```
albertshau: a page entry should not require a string representation of a record. It seems like this is being done in order to pass something to the transform method.

This iterator should not return PageEntry objects, but Page objects. The transform method should get Page objects from the InputFormat and call a method that will iterate through the records contained in the Page, emitting errors if needed. The PageEntry abstraction can be removed.

aonischuk: > a page entry should not require a string representation of a record. 
I use these text entries which failed to transform to structured record as a part of "send to error" routine. They are basically the body of InvalidEntry. 
This is useful so user can let's say handle his invalid json entries, which failed due to types or missing field.

src/main/java/io/cdap/plugin/http/source/common/pagination/entry/PageEntry.java
aonischuk: Done. Removed that class.

src/main/java/io/cdap/plugin/http/source/common/pagination/BaseHttpPaginationIterator.java
aonischuk: Done.

src/main/java/io/cdap/plugin/http/source/common/pagination/entry/PageEntry.java
aonischuk: Done.

Removed it from base class. Now it's only supplied with TransformationException for some formats (all except for BLOB).

Info from TransformationException is used to create InvalidEntry and emit it to error port.
This entry contains the information regarding the line which failed to transform.

This is pretty useful I would think. Assume that some json entries does not have some field or has a wrong type of it. The user will receive a text representation of those json entries from error port. The same goes for csv lines, xml nodes, etc.

aonischuk:
@albertshau addressed all comment. Can we merge If looks good?

src/main/java/io/cdap/plugin/http/source/batch/HttpBatchSource.java
```diff
@@ -92,55 +90,68 @@ public void prepareRun(BatchSourceContext context) {
...
-          throw new UnexpectedFormatException(
-            String.format("Unknown error handling strategy '%s'", config.getErrorHandling()));
+  public void transform(KeyValue<NullWritable, BasePage> input, Emitter<StructuredRecord> emitter) {
+    BasePage page = input.getValue();
+
+    if (handleHttpError(page, emitter)) {
+      return;
+    }
+
+    while (page.hasNext()) {
```
albertshau, line 94: It's not normal for an iterator to throw exceptions.

Instead of iterating in this method, it's better if BasePage has an emitRecords() method that takes an Emitter, with the Page class responsible for reading through the response and emitting output or errors. This puts the error handling in one place rather than some here and some in the Page classes.

src/main/java/io/cdap/plugin/http/source/common/http/HttpResponse.java
```diff
@@ -0,0 +1,89 @@
...
+  public InputStream getInputStream() throws IOException {
+    if (bytes != null) { // CloseableHttpResponse already read and closed the stream. So we need to use existing body.
+      return new ByteArrayInputStream(bytes);
+    } else {
+      HttpEntity responseEntity = response.getEntity();
+      return responseEntity.getContent();
+    }
+  }
+
+  public String getBody() {
```
albertshau, line 68: I don't think this is a clean API since it's expected that the caller will call either getInputStream() or getBody() or getBytes(). It's probably better for the Page classes just to get the CloseableHttpResponse like they were getting before, with each classes reading the response in the way that makes sense for it.

aonischuk: CloseableHttpResponse closes the response when read. So it cannot be read second time. It can pose a problem sometimes, let's say:
1. We read body first time to pass into custom pagination python code as argument and second time for text format to emit records.
2. We read the body in let's say JsonPage. But than error happens and we want to also emit the body as part of error and need to read it for the second time.
3. Any changes in future, where somebody will add reading body for the second time will fail. This class protects from any future failures like that.

Previously we just read one string in the beginning, and used it everywhere so problem was not present. But now as we don't want to save it all in memory this is the best / most transparent solution I see. Please let me know if you disagree.

src/main/java/io/cdap/plugin/http/source/batch/HttpBatchSource.java
aonischuk: I will re-use page logic for streaming plugin where there is no emitter. That's why I didn't want to pass emitter into common page classes.

src/main/java/io/cdap/plugin/http/source/common/http/HttpResponse.java
albertshau: This class does not protect the caller though. For example, if somebody calls getInputStream() and then getBody(), won't that result in an error? This class tries to hide the fact that you can't read the response twice, but there is no way to provide that abstraction without storing the entire contents.

src/main/java/io/cdap/plugin/http/source/batch/HttpBatchSource.java
albertshau: If this logic is going to be re-used, it's even more important for it to be in some common place rather than in the transform method. You don't have to pass in an emitter then, but can use something more generic like:

```
void consume(Consumer<StructuredRecord> onOutput, Consumer<InvalidEntry<StructuredRecord>> onError);
```

src/main/java/io/cdap/plugin/http/source/common/http/HttpResponse.java
aonischuk: Generally getInputStream() is called only by BasePage children. Which is the last step of the process. Should we maybe throw an exception if getBody()/getBytes() is called after getInputStream()? (which should never happen in our case)

The wrapper protects from other situations like "We read body first time to pass it into custom pagination python code as an argument and second time for text format to read it as stream.". Since one time it's read by pagination iterator (custom) and second time in TextPage it's really not a simple situation.

I'm not sure how we can overcome situations like this (there are also the same situation with errors). Without saving the whole string in wrapper, if it was already read.

albertshau: Hm, there must be a better way to structure this, but I suppose this is fine for now. Please add javadocs so the expected usage is clear.

src/main/java/io/cdap/plugin/http/source/batch/HttpBatchSource.java
aonischuk: Done

src/main/java/io/cdap/plugin/http/source/common/pagination/page/BasePage.java
```diff
@@ -42,14 +56,78 @@ public BasePage(BaseHttpSourceConfig config, HttpErrorHandler httpErrorHandler,
...
+  /**
+   * Run functions for every record on the page if appropriate. Or for whole page if has erroneous http
+   * status code.
+   *
+   * @param onOutput is run for non-erroneous records on the page
+   * @param onSendToError is run for erroneous records with "send to error" strategy. Or once for whole page with
+   *                      erroneous http status code and strategy "send to error".
+   */
+  public void consume(Consumer<StructuredRecord> onOutput, Consumer<InvalidEntry<StructuredRecord>> onSendToError)
+    throws TransformationException {
```
albertshau: this should no longer throw TransformationException

src/main/java/io/cdap/plugin/http/source/common/pagination/page/BasePage.java
```diff
@@ -42,14 +56,78 @@ public BasePage(BaseHttpSourceConfig config, HttpErrorHandler httpErrorHandler,
...
+            throw ex;
+          default:
+            throw new UnexpectedFormatException(
+              String.format("Unknown error handling strategy '%s'", config.getErrorHandling()));
+        }
+      }
+    }
+  }
+
+  private boolean handleHttpError(Consumer<InvalidEntry<StructuredRecord>> onSendToError) {
```
albertshau: Instead of having this logic here, its cleaner if there is an ErrorPage class instead of having this logic mixed in the base class. ErrorPage would emit an InvalidEntry in it's consume() method. The page iterator should create the page if the request failed and it is configured to send to error.

This separates the logic more cleanly and prevents somebody from overwriting the consume() method and forgetting to handle errors.

src/main/java/io/cdap/plugin/http/source/common/pagination/page/BasePage.java
```diff
@@ -42,14 +56,78 @@ public BasePage(BaseHttpSourceConfig config, HttpErrorHandler httpErrorHandler,
 
   @Nullable
   public abstract String getPrimitiveByPath(String path);
+  public abstract boolean hasNext();
+  protected abstract StructuredRecord next() throws TransformationException;
```
albertshau: TransformationException here is being used like a return value. It's better if the actual implementation creates the InvalidEntry and calls the consumer rather than using the exception to control the flow of the logic. Different implementations may also want to include different fields in the error.

I think it's simpler to make consume() abstract and have the children implement it. It would also make the logic in things like RecordPerLinePage simpler, as reading through an InputStream is more straightforward than implementing an iterator over a stream.

aonischuk: Unfortunately we need Page to be an iterator.

The reason is that in BasePaginationIterator we need to check page.hasNext(). And if the page does not have elements the pagination iterator needs to stop loading pages.

Context: A lot of paginated APIs return endless amount of empty pages once all elements run out.

src/main/java/io/cdap/plugin/http/source/common/pagination/page/BasePage.java
aonischuk: Since we need Page to be an iterator (read the comment below, where I explained it). I don't think we can implement it like this with BasePage still being an iterator.

src/main/java/io/cdap/plugin/http/source/common/pagination/page/BasePage.java
aonischuk: In case the error handling strategy is "stop on error" it will throw TransformationException with detailed information due to which transformation it failed.

That's why we need this throws.

aonischuk:
@albertshau while implementing Realtime plugin I have stumbled uppon the fact the JavaStreamingContext#flatMap requires an Iterator to be returned in order to turn a value into multiple values in scope of spark stream (BasePage into set of StructuredRecords).

Due to those circumstances I had to make a BasePage an iterator and remove "consume" method in order for it to work, and also make next() not throw TransformationException, but instead handle it inside the next() method itself.

__Can we please go back to BasePage being an iterator in order to effectively implement this?__

Ps. I haven't found a way to return structuredRecords to stream from a function (consume callback).

Also one more thing handleHttpError in streaming has to be run in custom receiver (not parallel part), while iteration has to happen in a stream (parallel part), that's why we also need them to be called separately (not in scope of one consume call)

albertshau:
It is ok if it behaves like a normal Iterator, meaning it doesn't throw exceptions that the caller is expected to handle. Errors are expected behavior that should be handled as close to the source of the error as possible.

aonischuk:
@albertshau done. Can you please re-review?

src/main/java/io/cdap/plugin/http/source/common/BaseHttpSourceConfig.java
```diff
@@ -25,7 +25,7 @@
 import io.cdap.plugin.common.ReferencePluginConfig;
 import io.cdap.plugin.http.source.common.error.ErrorHandling;
 import io.cdap.plugin.http.source.common.error.HttpErrorHandlerEntity;
-import io.cdap.plugin.http.source.common.error.HttpErrorHandlingStrategy;
+import io.cdap.plugin.http.source.common.error.RetriableErrorHandling;
```
albertshau: nit: elsewhere in the CDAP codebase we've spelled this as Retryable. I don't think there's a canonical way to spell it, but let's use Retryable instead of Retriable for consistency.

albertshau (approved):
minor comment about renaming a class, otherwise lgtm. please fix and squash, then i'll merge.

src/main/java/io/cdap/plugin/http/source/common/BaseHttpSourceConfig.java
aonischuk: Done. Once this is merged can you please also merge #13 as well.

PR #13, opened 2019-07-23
CDAP-15524. HTTP Batch Source integration tests

aonischuk:
Integration tests use wireMock as http server mock. Which is really simple to use in tests to mock http calls.


albertshau (approved):
lgtm, though will wait for other PR before merging

PR #14, opened 2019-08-09
CDAP-15672. HTTP Realtime Source

aonischuk:
Implement Realtime source. Which supports:

- checkpointing. To continue from the last processed page
- uses InputDStream for checkpointing
- returns RDD from inputDStream by parts
- supports comparing and reloading the last page records if the page changes (hash checks). Is controlled by "Track last page changes" config.

For more details: https://wiki.cask.co/display/CE/HTTP+Realtime+Source

### DIFF
--- a/docs/HTTP-streamingsource.md
+++ b/docs/HTTP-streamingsource.md
@@ -1,0 +1,426 @@
+# HTTP Streaming Source
+
+Description
+-----------
+This plugin reads data from HTTP/HTTPS pages periodically.
+Paginated APIs are supported. For paginated APIs plugin reads available data and
+than waits for new pages to appear.
+Data in JSON, XML, CSV, TSV, TEXT and BLOB formats is supported.
+
+Properties
+----------
+
+### General
+
+**Reference Name:** Name used to uniquely identify this source for lineage, annotating metadata, etc.
+
+**URL:** Url to fetch to the first page.
+The url must start with a protocol (e.g. http://).
+
+**HTTP Method:** HTTP request method.
+
+**Headers:** Headers to send with each HTTP request.
+
+**Request body:** Body to send with each HTTP request.
+
+**Max Pages Per Fetch:** Maximum number of pages put to RDD in one blocking reading. Empty value means that the
+maximum is not enforced.
+
+### Format
+
+**Format:** Format of the HTTP response. This determines how the response is converted into output records. Possible values are:<br>
+JSON - retrieves all records from the given json path
+and transforms them into records according to the mapping.<br>
+XML - retrieves all records from the given XPath
+and transforms them into records according to the mapping.<br>
+TSV - tab separated values. Columns are mapped to record fields in the order they are
+listed in schema.<br>
+CSV - comma separated values. Columns are mapped to record fields in the order they are
+listed in schema.<br>
+Text - transforms a single line of text into a single record with a string field `body` containing the result.<br>
+BLOB - transforms the entire response into a single record with a byte array field `body` containing the result.
+
+**JSON/XML Result Path:** Path to the results. When the format is XML, this is an XPath. When the format is JSON, this is a JSON path.
+
+JSON path example:
+```
+{
+     "errors": [],
+     "response": {
+       "books": [
+         {
+           "id": "1159142",
+           "title": "Agile Web Development with Rails",
+           "author": "Sam Ruby, Dave Thomas, David Heinemeier Hansson",
+           "printInfo": {
+             "page": 488,
+             "coverType": "hard",
+             "publisher": "Pragmatic Bookshelf"
+           }
+         },
+         {
+           "id": "2375753",
+           "title": "Flask Web Development",
+           "author": "Miguel Grinberg",
+           "printInfo": {
+             "page": 543,
+             "coverType": "hard",
+             "publisher": "O'Reilly Media, Inc"
+           }
+         },
+         {
+           "id": "547307",
+           "title": "Alex Homer, ASP.NET 2.0 Visual Web Developer 2005",
+           "author": "David Sussman",
+           "printInfo": {
+             "page": 543,
+             "coverType": "hard",
+             "publisher": "unknown"
+           }
+         }
+       ]
+     }
+}
+ ```
+Json path to fetch books is `/response/books`. However if we need to fetch only `printInfo` we can specify
+`/response/books/printInfo` as well.
+
+XPath example:
+```
+<?xml version="1.0" encoding="UTF-8"?>
+<bookstores>
+  <bookstore id="1">
+     <book category="cooking">
+        <title lang="en">Everyday Italian</title>
+        <author>Giada De Laurentiis</author>
+        <year>2005</year>
+        <price>
+         <value>15.0</value>
+         <policy>Discount up to 50%</policy>
+        </price>
+     </book>
+     <book category="web">
+        <title lang="en">XQuery Kick Start</title>
+        <author>James McGovern</author>
+        <author>Per Bothner</author>
+        <year>2003</year>
+        <price>
+         <value>49.99</value>
+         <policy>No discount</policy>
+        </price>
+     </book>
+     ...
+  </bookstore>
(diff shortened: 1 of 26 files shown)


src/main/java/io/cdap/plugin/http/source/common/pagination/IncrementAnIndexPaginationIterator.java
```diff
@@ -61,4 +67,28 @@ protected String getNextPageUrl(HttpResponse response, BasePage page) {
...
   }
+
+  @Override
+  public void setNextPageUrl(String nextPageUrl) {
+    super.setNextPageUrl(nextPageUrl);
+
+    Matcher matcher = urlPatten.matcher(nextPageUrl);
+    if (!matcher.matches()) {
+      throw new IllegalArgumentException(String.format(
+        "Cannot set nextPageUrl to '%s' as it does not match the original url '%s'", nextPageUrl, config.getUrl()));
```
nitinmotgi: Simple English text no nextPageUrl variable names. Refer to display names

pom.xml
```diff
@@ -240,6 +241,71 @@
       </exclusions>
     </dependency>
 
+    <dependency>
+      <groupId>io.cdap.cdap</groupId>
+      <artifactId>cdap-etl-api-spark</artifactId>
+      <version>${cdap.version}</version>
```
albertshau, line 247: should be provided scope

src/main/java/io/cdap/plugin/http/source/streaming/HttpInputDStream.java
```diff
@@ -0,0 +1,141 @@
...
+      return doCompute();
+    } catch (Exception e) {
+      throw Throwables.propagate(e);
+    }
+  }
+
+  private Option<RDD<StructuredRecord>> doCompute() throws IOException {
+    try (BaseHttpPaginationIterator httpPaginationIterator = PaginationIteratorFactory.createInstance(config)) {
+      if (this.lastProcessedPageUrl != null) { // nothing was processed before
+        httpPaginationIterator.setNextPageUrl(this.lastProcessedPageUrl);
```
albertshau: It wasn't clear to me what the contract for this method was until reading this class. This is kind of an inconvenient API, as it requires implementations to be able to extract all their required information from a url using pattern matching, which is more complex and error prone than ideal.

Instead of using the current page url as state, I think it would be better to have a separate state object that gets tracked by the InputDStream. When we create the pagination iterator, we pass in the state to the createInstance() method. Most iterators may decide to use the url as their state, but others can choose to use something else. For example, the increment iterator would just use the index. To start, state could just be a String.

src/main/java/io/cdap/plugin/http/source/streaming/HttpInputDStream.java
```diff
@@ -0,0 +1,141 @@
...
+
+      try {
+        while (httpPaginationIterator.hasNext()) {
+          BasePage page = httpPaginationIterator.next();
+          String pageUrl = httpPaginationIterator.getCurrentPageUrl();
+          LOG.debug("Visited {}", pageUrl);
+
+          // skip last page, if we don't track last page changes or if previously processed page had the same hash
+          if (pageUrl.equals(this.lastProcessedPageUrl) &&
+            (!config.shouldTrackLastPageChanges() || page.getHash() == lastProcessedPageHash)) {
```
albertshau: I don't think we should ever need to skip a page as long as the PaginationIterator that is created starts on the page after the last processed page. When a pipeline is stopped, it will tell Spark to stop gracefully, meaning it will stop calling compute() after the current batch interval and wait for the microbatches to be processed downstream. 

So the problem is only when the pipeline can't stop gracefully, but in that scenario storing the hash and sometimes skipping the page won't prevent data from getting skipped either.

widgets/HTTP-streamingsource.json
```diff
@@ -0,0 +1,525 @@
...
+        {
+          "widget-type": "textarea",
+          "name": "requestBody",
+          "label": "Request Body",
+          "widget-attributes": {
+            "rows": "5"
+          }
+        },
+        {
+          "widget-type": "radio-group",
```
albertshau: boolean properties should use the 'toggle' widget type

widgets/HTTP-streamingsource.json
```diff
@@ -0,0 +1,525 @@
...
+          "label": "Track Last Page Changes",
```
albertshau: Track Page State

docs/HTTP-streamingsource.md
```diff
@@ -0,0 +1,425 @@
...
+**Linear Retry Interval:** Interval between retries. Is only used if retry policy is "linear".
+
+**Max Retry Duration:** Maximum time in seconds retries can take.
+
+**Connect Timeout:** Maximum time in seconds connection initialization is allowed to take.
+
+**Read Timeout:** Maximum time in seconds fetching data from the server is allowed to take.
+
+### Pagination
+
```
albertshau, line 260: There should be some general information here about how the plugin behaves when there is pagination vs when there isn't. 

When there is pagination, it reads pages in order, keeping track of which pages have already been read. It will not reprocess any page it has already read.

When there is no pagination, it periodically polls the same page.

docs/HTTP-streamingsource.md
```diff
@@ -0,0 +1,425 @@
...
+**URL:** Url to fetch to the first page.
+The url must start with a protocol (e.g. http://).
+
+**HTTP Method:** HTTP request method.
+
+**Headers:** Headers to send with each HTTP request.
+
+**Request body:** Body to send with each HTTP request.
+
+**Track Last Page Changes:** If true, until new page appears, plugin will track the last page available and if
```
albertshau: I don't think we need to support this in the initial version. Not sure how useful it is as it only affects things if there is no pagination, but it doesn't prevent duplicate records, only duplicate pages.

src/main/java/io/cdap/plugin/http/source/streaming/HttpInputDStream.java
aonischuk: Actually we have to store the last page url which had results and skip it every time when waiting for a new page.

The reason is that in a bunch of pagination iterators (e.g. pagination by header, pagination by field in body, pagination by token in body), if there is no next page the page won't contain nextPageUrl and so we will need to poll it everytime and skip it until the nextPageUrl appears on it.

Ps. Hash was only used for "Track Last Page Changes" = true, but now removed it's use.

src/main/java/io/cdap/plugin/http/source/common/pagination/IncrementAnIndexPaginationIterator.java
aonischuk: Done. Actually code with exception got removed due to changes suggested by Albert.

pom.xml
aonischuk: Done

src/main/java/io/cdap/plugin/http/source/streaming/HttpInputDStream.java
aonischuk: Done

docs/HTTP-streamingsource.md
aonischuk: Done. Removed "Track Last Page Changes" option.

docs/HTTP-streamingsource.md
aonischuk: Done

widgets/HTTP-streamingsource.json
aonischuk: Done. Removed "Track Last Page Changes" option.

widgets/HTTP-streamingsource.json
aonischuk: Done. Removed "Track Last Page Changes" option.

albertshau (approved):
lgtm, please squash commits and i'll merge

aonischuk:
@albertshau squashed the commits. Can we please merge?

PR #15, opened 2019-08-15
CDAP-15672. HTTP streaming source integration tests.

aonischuk:
Make already existing batch integration tests class an abstract class. And extend it by both HttpStreamingSourceETLTest and HttpBatchSourceETLTest.


aonischuk:
@albertshau can you please review this?

Ps. 
code from the HttpSourceETLTest.java is mostly moved from HttpBatchSourceETLTest.java. Not a new code 90%.
So the actual changes are smaller here. Than looks.

src/test/java/io/cdap/plugin/http/etl/HttpSourceETLTest.java
```diff
@@ -0,0 +1,518 @@
...
+import org.junit.rules.TestName;
+
+import java.io.IOException;
+import java.net.URISyntaxException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.List;
+import java.util.Map;
+
+public abstract class HttpSourceETLTest extends HydratorTestBase {
```
albertshau, line 39: Can you comment on which part is different and should be reviewed?

aonischuk: I have separated the changes into two commits. So git recognizes what was moved. Now the review should be much easier.

aonischuk: So please only review the second commit. Instead of "Files changed" all together.

albertshau (approved):

PR #16, opened 2019-09-16
[CDAP-15919] Adds Dynamic plugin filters to http plugin

ajainarayanan:


widgets/HTTP-batchsource.json
```diff
@@ -101,6 +102,60 @@
         }
       ]
     },
+    {
+      "label": "OAuth2",
+      "properties": [
+        {
+          "widget-type": "radio-group",
```
albertshau: existing, but for new plugins I think we should use the toggle widget for booleans.

elfenheart: done

widgets/HTTP-streamingsource.json
```diff
@@ -502,5 +497,188 @@
...
+        {
+          "name": "refreshToken",
+          "type": "property"
+        }
+      ]
+    },
+    {
+      "name": "JSON/XML Formatting",
+      "condition": {
+        "expression": "format == 'json' || format == 'xml'"
```
bdmogal, line 656: I was going to open a jira about this. Shouldn't we instead support operators such as 'in', 'not in', 'matches', 'not matches', 'contains', 'not contains'? For simple conditions like these, we shouldn't need to specify an expression right?

@elfenheart @albertshau

albertshau: +1. I think the workaround today would be to have two filters, one for json and one for xml.

albertshau (approved):

widgets/HTTP-streamingsource.json
bdmogal: https://issues.cask.co/browse/CDAP-16023

ajainarayanan: This is right now possible. https://github.com/TomFrost/Jexl#a-note-about-in

PR #21, opened 2019-11-19
Bump version on develop

CuriousVini:


pom.xml
```diff
@@ -567,8 +567,8 @@
         <version>1.1.0</version>
         <configuration>
           <cdapArtifacts>
-            <parent>system:cdap-data-pipeline[6.1.0-SNAPSHOT,7.0.0-SNAPSHOT)</parent>
-            <parent>system:cdap-data-streams[6.1-SNAPSHOT.0,7.0.0-SNAPSHOT)</parent>
+            <parent>system:cdap-data-pipeline[6.2.0-SNAPSHOT,7.0.0-SNAPSHOT)</parent>
```
mikkin: Does this mean that until we release 6.1.0, we won't be able to build and run plugin in develop?

mikkin (commented):
One minor question, but otherwise looks good

pom.xml
CuriousVini: I think we can modify it to 6.2.0-SNAPSHOT if we add new apis in 6.2.0-SNAPSHOT. Changing it to 6.2.0-SNAPSHOT

PR #31, opened 2020-04-16
bumped version

bdmogal:


albertshau (approved):

albertshau:
it would also be good to set the cdap version to 6.1.1, unless it really is using a new feature from 6.2.0-SNAPSHOT

PR #32, opened 2020-04-17
(PLUGIN-131) Fixed source to not error out if it can't find an option…

bdmogal:
…al/nullable field

Jira: [PLUGIN-131](https://issues.cask.co/browse/PLUGIN-131)

Summary:
The plugin was causing pipelines to fail even if some optional fields in the schema were missing in the response from the API. Instead of failing the pipeline always, this change now does two main things:
1. If the missing field is nullable, ignores it.
2. If the missing field is not nullable, uses the error handling option that users have specified in the configuration to treat it as a data error, and doesn't fail the pipeline.

### DIFF
--- a/src/main/java/io/cdap/plugin/http/source/common/pagination/page/JSONUtil.java
+++ b/src/main/java/io/cdap/plugin/http/source/common/pagination/page/JSONUtil.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2019 Cask Data, Inc.
+ * Copyright © 2019-2020 Cask Data, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
@@ -15,6 +15,7 @@
  */
 package io.cdap.plugin.http.source.common.pagination.page;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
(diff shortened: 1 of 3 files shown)


bdmogal:
tested it with https://cat-fact.herokuapp.com/facts, which was failing prior to this change since the field `user` was missing in some records.

src/main/java/io/cdap/plugin/http/source/common/pagination/page/JSONUtil.java
```diff
@@ -84,17 +85,18 @@ public static JsonQueryResponse getJsonElementByPath(JsonObject jsonObject, Stri
...
     private final JsonElement result;
 
-    public JsonQueryResponse(String[] retrievedPathParts, String[] unretrievedPathParts, JsonElement result) {
-      this.retrievedPathParts = retrievedPathParts;
+    private JsonQueryResponse(String[] retrievedPathParts, String[] unretrievedPathParts,
+                              List<String> optionalPathParts, JsonElement result) {
       this.unretrievedPathParts = unretrievedPathParts;
       this.retrievedPath = "/" + StringUtils.join(retrievedPathParts, '/');
       this.unretrievedPath = "/" + StringUtils.join(unretrievedPathParts, '/');
+      this.optionalPathParts = optionalPathParts;
```
chtyim: Is this field or path? the caller calls this field, while it is called path here.

bdmogal: This class represents a parsed JSON, and does it by using some JSON Path (format ``/path/to/field``. However, I am passing option fields from the schema in here, so changed it to ``optionalFields``

chtyim (approved):

PR #44, opened 2020-12-04
PLUGIN-123: Fixed http source plugin, the source now supports retrieving data as list

flakrimjusufi:
HTTP Source should support APIs that return a list at the top level

JIRA ticket: https://cdap.atlassian.net/browse/PLUGIN-123

### DIFF
--- a/src/main/java/io/cdap/plugin/http/source/common/BaseHttpSourceConfig.java
+++ b/src/main/java/io/cdap/plugin/http/source/common/BaseHttpSourceConfig.java
@@ -774,7 +774,9 @@ public abstract class BaseHttpSourceConfig extends ReferencePluginConfig {
       String reasonFormat = String.format("page format is '%s'", getFormat());
 
       if (getFormat().equals(PageFormat.JSON) || getFormat().equals(PageFormat.XML)) {
-        assertIsSet(getResultPath(), PROPERTY_RESULT_PATH, reasonFormat);
+        if (!getFormat().equals(PageFormat.JSON)) {
+          assertIsSet(getResultPath(), PROPERTY_RESULT_PATH, reasonFormat);
+        }
         getFullFieldsMapping(); // can be null, but call getter to verify correctness of regexps
       } else {
         assertIsNotSet(getResultPath(), PROPERTY_RESULT_PATH, reasonFormat);
--- a/src/main/java/io/cdap/plugin/http/source/common/pagination/page/JSONUtil.java
+++ b/src/main/java/io/cdap/plugin/http/source/common/pagination/page/JSONUtil.java
@@ -25,6 +25,7 @@ import org.apache.commons.lang3.StringUtils;
 
 import java.util.Arrays;
 import java.util.List;
+import javax.annotation.Nullable;
 
 /**
  * Utility functions for working with JSON document.
@@ -34,6 +35,10 @@ public class JSONUtil {
 
   public static JsonObject toJsonObject(String text) {
     return JSON_PARSER.parse(text).getAsJsonObject();
+  }
+
+  public static JsonElement toJsonElement(String text) {
+    return JSON_PARSER.parse(text);
   }
 
   public static JsonArray toJsonArray(String text) {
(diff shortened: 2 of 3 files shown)


src/main/java/io/cdap/plugin/http/source/common/pagination/page/JSONUtil.java
```diff
@@ -51,8 +55,11 @@ public static JsonArray toJsonArray(String text) {
    */
   public static JsonQueryResponse getJsonElementByPath(JsonObject jsonObject, String jsonPath,
```
rmstar: Annotate jsonPath as Nullable

src/main/java/io/cdap/plugin/http/source/common/BaseHttpSourceConfig.java
```diff
@@ -774,7 +774,9 @@ public void validate() {
       String reasonFormat = String.format("page format is '%s'", getFormat());
 
       if (getFormat().equals(PageFormat.JSON) || getFormat().equals(PageFormat.XML)) {
-        assertIsSet(getResultPath(), PROPERTY_RESULT_PATH, reasonFormat);
+        if (!getFormat().equals(PageFormat.JSON)) {
+          assertIsSet(getResultPath(), PROPERTY_RESULT_PATH, reasonFormat);
```
rmstar, line 778: In what cases can resultPath be unset?

rmstar (commented):
Can you please list the test cases you ran to validate these changes?

From the JIRA description:
```
 if it returns an array of objects, the source should return multiple records, one for each object in the array. 
Schema should be set based on this logic, and pagination should also work similarly.
 ```
 Is this handled in this PR? If yes, can you please point me to the part of PR that handles splitting into multiple records/pagination? Thanks.

**Comment by flakrimjusufi** [comment]
@rmstar 

Here you have the test cases I've used to validate these changes: 

**1. Simple JSON with path** 
```
{
	"result": {
		"username": "kimi",
		"host": "localhost",
		"port": "88"
	}
}

```
**2. JSON array without a path** 
```
[
	{
		"username": "kimi",
		"host": "localhost",
		"port": "88"
	}
]

```
**3. JSON nested object with two levels and path**
```
{
	"result" : {
			"body":{
				"username" : "kimi",
				"host": "localhost",
				"port": "88"
			}
		}
}

```

**4. JSON array with two objects and path** 

```
[
   {
      "result":{
         "username":"kimi",
         "host":"localhost",
         "port":"88"
      }
   },
   {
      "result":{
         "username":"ardi",
         "host":"localhost",
         "port":"80"
      }
   }
   
]
```

src/main/java/io/cdap/plugin/http/source/common/BaseHttpSourceConfig.java
flakrimjusufi: resultPath can be unset in the cases where the top level is just a JSONArray without a path. 



> Can you please list the test cases you ran to validate these changes?
> 
> From the JIRA description:
> 
> ```
>  if it returns an array of objects, the source should return multiple records, one for each object in the array. 
> Schema should be set based on this logic, and pagination should also work similarly.
> ```
> 
> Is this handled in this PR? If yes, can you please point me to the part of PR that handles splitting into multiple records/pagination? Thanks.


The pagination will work just fine when the top level will be JSONObject with path included. 

Here you have the part of the PR that handles the pagination: 
https://github.com/data-integrations/http/blob/67b752d46aa01f0171a1e90f5ff30a37738897a9/src/main/java/io/cdap/plugin/http/source/common/pagination/page/JsonPage.java#L196-L205

In the cases where the top level is _JSONArray_, the pagination for types such as **Link in response body** and **Token in response body** makes no sense for me. 

Please correct me if my assumption is wrong.

rmstar: I see, thanks

rmstar (approved):
Please squash your commits

flakrimjusufi:
> Please squash your commits

Done.

rmstar (approved):

PR #51, opened 2021-03-11
PLUGIN-219: Pipeline with HttpSink plugin fails to deploy if URL field is specified as macro.

Roni98:
Pipeline with HttpSink plugin fails to deploy if URL field is specified as macro.
JIRA ticket: https://cdap.atlassian.net/browse/PLUGIN-219

### DIFF
--- a/src/test/java/io/cdap/plugin/http/sink/batch/HTTPSinkConfigTest.java
+++ b/src/test/java/io/cdap/plugin/http/sink/batch/HTTPSinkConfigTest.java
@@ -17,6 +17,7 @@
 package io.cdap.plugin.http.sink.batch;
 
 import io.cdap.cdap.etl.api.validation.CauseAttributes;
+import io.cdap.cdap.etl.api.validation.ValidationException;
 import io.cdap.cdap.etl.api.validation.ValidationFailure;
 import io.cdap.cdap.etl.mock.validation.MockFailureCollector;
 import org.junit.Assert;
@@ -132,6 +133,17 @@ public class HTTPSinkConfigTest {
     MockFailureCollector failureCollector = new MockFailureCollector(MOCK_STAGE);
     config.validate(failureCollector);
     assertPropertyValidationFailed(failureCollector, HTTPSinkConfig.MESSAGE_FORMAT);
+  }
+
+  @Test(expected = ValidationException.class)
+  public void testHTTPSinkWithEmptyUrl() {
+    HTTPSinkConfig config = HTTPSinkConfig.newBuilder(VALID_CONFIG)
+      .setUrl("")
+      .build();
+
+    MockFailureCollector collector = new MockFailureCollector("httpsinkwithemptyurl");
+    config.validate(collector);
+    collector.getOrThrowException();
   }
 
   public static void assertPropertyValidationFailed(MockFailureCollector failureCollector, String paramName) {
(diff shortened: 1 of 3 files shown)


src/test/java/io/cdap/plugin/http/sink/batch/HTTPSinkConfigTest.java
```diff
@@ -67,6 +75,32 @@ public void testInvalidUrl() {
...
+    PluginProperties rawProperties = PluginProperties.builder()
+      .addAll(properties)
+      .build()
+      .setMacros(macros);
+
+    FieldSetter fs = new FieldSetter(config, HTTPSinkConfig.class.getSuperclass().getSuperclass()
+      .getDeclaredField("properties"));
+    fs.set(rawProperties);
+
+
```
flakrimjusufi, line 98: nit: Move one line up.

src/test/java/io/cdap/plugin/http/sink/batch/HTTPSinkConfigTest.java
```diff
@@ -67,6 +75,32 @@ public void testInvalidUrl() {
     assertPropertyValidationFailed(failureCollector, HTTPSinkConfig.URL);
   }
 
+  @Test
+  public void testMacroUrl() throws Exception {
```
flakrimjusufi: This is not the approach on how we test with Macro. 

Please take this as a reference and change your test accordingly:

https://github.com/cdapio/hydrator-plugins/blob/29a16d6e9f70447144930be47b85a37c99c1cc87/core-plugins/src/test/java/io/cdap/plugin/batch/sink/TableSinkTest.java#L255

flakrimjusufi (changes requested):

src/test/java/io/cdap/plugin/http/sink/batch/HTTPSinkConfigTest.java
Roni98: PR updated.

flakrimjusufi (commented):
It looks good to me. 

Now you can go into verification process.

src/test/java/io/cdap/plugin/http/sink/batch/HTTPSinkTest.java
```diff
@@ -174,4 +177,69 @@ private int getFeeds() throws IOException {
     urlConn.disconnect();
     return responseCode;
   }
+
+  @Test
+  public void testHTTPSinkMacroUrl() throws Exception {
+    String inputDatasetName = "input-http-sink-with-macro";
+    ETLStage source = new ETLStage("source", MockSource.getPlugin(inputDatasetName));
+    Map<String, String> properties = new ImmutableMap.Builder<String, String>()
+      .put("url", "${url}")
```
CuriousVini, line 186: please add a negative test when url is null or empty

Roni98: PR updated.

src/test/java/io/cdap/plugin/http/sink/batch/HTTPSinkTest.java
```diff
@@ -174,4 +176,109 @@ private int getFeeds() throws IOException {
...
+      StructuredRecord.builder(inputSchema).set("id", "1").build()
+    );
+    MockSource.writeInput(inputManager, input);
+    // run the pipeline
+    runETLOnce(appManager, runtimeProperties);
+    verifyandcleanupTest();
+  }
+
+  @Test
+  public void testHTTPSinkWithEmptyUrl() {
```
CuriousVini: If there is exception that is thrown we should rely on junit for that. We should not silently ignore the exception
`@Test(expected=<SomeException>.class)`

Roni98: PR updated accordingly

CuriousVini (approved):
please squash the commits

PR #57, opened 2021-05-07
Add Hawk Authentication for sources

VictorD-Veolia:
Integration of [HAWK authentication](https://github.com/mozilla/hawk) for HTTP sources

### DIFF
--- a/docs/HTTP-batchsource.md
+++ b/docs/HTTP-batchsource.md
@@ -390,6 +390,24 @@ is stopped.
 
 **Refresh Token:** Token used to receive accessToken, which is end product of OAuth2.
 
+### Hawk Authentication
+
+**HAWK Authentication Enabled:** If true, plugin will perform HAWK authentication.
+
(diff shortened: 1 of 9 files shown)


VictorD-Veolia:
@googlebot I signed it!

bdmogal:
@VictorD-Veolia thanks for the PR. Could you share a short document describing the use case for this, since I'm not familiar with HAWK?

VictorD-Veolia:
Hi @bdmogal, HAWK is an HTTP authentication scheme providing mechanisms for making authenticated HTTP requests with partial cryptographic verification of the request and response, covering the HTTP method, request URI, host, and optionally the request payload.

Some teams may prefer HAWK authentication over Oauth2 for their APIs as a more secure option when additional verification must be done on requests and responses.

avirmani:
@VictorD-Veolia - is there a small 1-pager that you can share to explain this a bit more, pls? Also, we should look at adding some unit tests here?

PR #58, opened 2021-05-07
Fix nullable fields handling with JSON source 

VictorD-Veolia:
Fix an error in JSONPage that prevented HTTP sources to handle nullable fields correctly.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -82,7 +82,7 @@
     <gson.version>2.8.5</gson.version>
     <hadoop.version>2.3.0</hadoop.version>
     <httpcomponents.version>4.5.9</httpcomponents.version>
-    <hydrator.version>2.4.0-SNAPSHOT</hydrator.version>
+    <hydrator.version>2.4.0</hydrator.version>
     <jackson.version>2.9.9</jackson.version>
     <junit.version>4.11</junit.version>
     <jython.version>2.7.1</jython.version>
--- a/pom.xml
+++ b/pom.xml
@@ -82,7 +82,7 @@
     <gson.version>2.8.5</gson.version>
     <hadoop.version>2.3.0</hadoop.version>
     <httpcomponents.version>4.5.9</httpcomponents.version>
-    <hydrator.version>2.4.0</hydrator.version>
+    <hydrator.version>2.10.0-SNAPSHOT</hydrator.version>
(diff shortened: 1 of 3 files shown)


bdmogal:
@VictorD-Veolia do you mind sharing the exact issue (some sample response for which the plugin failed before the fix)? If you can add a test to demonstrate it, that would be even better.

bdmogal:
Also, since this repository does not have automatic builds set up yet, would it be possible to paste the output of `mvn clean package` with this change?

VictorD-Veolia:
Hi @bdmogal, I added a unit test and created an issue in JIRA : https://cdap.atlassian.net/browse/PLUGIN-805

I also added to this comment the output of the `mvn clean package` you asked.

[mvn_package.log](https://github.com/data-integrations/http/files/6974463/mvn_package.log)

pom.xml
```diff
@@ -82,7 +82,7 @@
     <gson.version>2.8.5</gson.version>
     <hadoop.version>2.3.0</hadoop.version>
     <httpcomponents.version>4.5.9</httpcomponents.version>
-    <hydrator.version>2.4.0-SNAPSHOT</hydrator.version>
+    <hydrator.version>2.4.0</hydrator.version>
```
fernst: We should bump to a release version in the develop branch

src/test/java/io/cdap/plugin/http/source/common/pagination/page/JSONPageTest.java
```diff
@@ -0,0 +1,109 @@
+/*
+ * Copyright © 2019-2020 Cask Data, Inc.
```
rmstar: 2022

pom.xml
```diff
@@ -82,7 +82,7 @@
     <gson.version>2.8.5</gson.version>
     <hadoop.version>2.3.0</hadoop.version>
     <httpcomponents.version>4.5.9</httpcomponents.version>
-    <hydrator.version>2.4.0-SNAPSHOT</hydrator.version>
+    <hydrator.version>2.10.0-SNAPSHOT</hydrator.version>
```
rmstar, line 85: This should be set to a non-snapshot version

rmstar (approved):
please squash commits after addressing comment

PR #60, opened 2021-05-11
add HTTP plugin support for service account authentication

ropeck:
The issue here is that the oauth provided in the plugin doesn't have any support for service account users. It has a set of config parameters to enter, including a refresh token. That token is used to request a fresh access token when doing oauth for users. For a service account, there's no way to make a refresh token. Instead, a service account uses a private key and other details that are given in the service account json keyfile. The customer needs a way to use the http plugin using that data for a service account. It's taking more work than planned, but I have code working for a custom http plugin adding auth with a service account json key. Using this, I made a test fusion instance pipeline that calls the GCE API to list my compute engine instances and then writes that data to a GCS bucket. Authentication is done with a service account json keyfile given in the configuration for the http plugin. I'll send an update to the customer to find out if this is all they need to unblock. There are several things still to implement after this proof of concept change to the plugin. I've updated the batch source plugin, but that leave the streaming source plugin and both stream and batch for the sink output plugin. The changes there are the same as for the batch source code.

The code took a long time to finish and test because I ran into a lot of dependency conflicts with guava in the pipeline code and the current versions that the google oauth packages expect. Doing Base64 parsing and fetching web page results failed and needed local workarounds. If there's some way to clean up the code and avoid those parts of the code, it would be much simpler to read.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -21,7 +21,7 @@
   <name>HTTP Plugins</name>
   <groupId>io.cdap</groupId>
   <artifactId>http-plugins</artifactId>
-  <version>1.4.0-SNAPSHOT</version>
+  <version>1.4.1-service-account</version>
 
   <licenses>
     <license>
@@ -82,7 +82,7 @@
     <gson.version>2.8.5</gson.version>
     <hadoop.version>2.3.0</hadoop.version>
     <httpcomponents.version>4.5.9</httpcomponents.version>
-    <hydrator.version>2.4.0-SNAPSHOT</hydrator.version>
+    <hydrator.version>2.3.0-SNAPSHOT</hydrator.version>
     <jackson.version>2.9.9</jackson.version>
     <junit.version>4.11</junit.version>
     <jython.version>2.7.1</jython.version>
@@ -93,6 +93,16 @@
   </properties>
 
   <dependencies>
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+      <version>13.0.1</version>
+      </dependency>
+       <dependency>
+         <groupId>com.google.auth</groupId>
+         <artifactId>google-auth-library-oauth2-http</artifactId>
+         <version>0.25.5</version>
+       </dependency>
     <dependency>
       <groupId>io.cdap.cdap</groupId>
       <artifactId>cdap-api</artifactId>
@@ -354,8 +364,6 @@
       <artifactId>jython-standalone</artifactId>
       <version>${jython.version}</version>
     </dependency>
-
-
     <!-- tests -->
     <dependency>
       <groupId>io.cdap.cdap</groupId>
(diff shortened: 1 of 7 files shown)


pom.xml
```diff
@@ -82,7 +82,7 @@
     <gson.version>2.8.5</gson.version>
     <hadoop.version>2.3.0</hadoop.version>
     <httpcomponents.version>4.5.9</httpcomponents.version>
-    <hydrator.version>2.4.0-SNAPSHOT</hydrator.version>
+    <hydrator.version>2.3.0-SNAPSHOT</hydrator.version>
```
avirmani, line 85: Why did we have to go to an older version?

pom.xml
```diff
@@ -21,7 +21,7 @@
   <name>HTTP Plugins</name>
   <groupId>io.cdap</groupId>
   <artifactId>http-plugins</artifactId>
-  <version>1.4.0-SNAPSHOT</version>
+  <version>1.4.1-service-account</version>
```
rmstar, line 24: Please revert the version change heree.  We'll bump the version on the release branch after cherrypicking your PR.

avirmani, line 24: When we publish in the hub we would change this, right?

pom.xml
rmstar, line 85: Why did we change this. Also, we probably shouldn't be depending on a snapshot version here.

pom.xml
```diff
@@ -93,6 +93,16 @@
   </properties>
 
   <dependencies>
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
```
avirmani, line 98: Do we know if this could break any other plugins due to conflicting dependencies?

src/main/java/io/cdap/plugin/http/source/common/http/OAuthUtil.java
```diff
@@ -15,22 +15,176 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 /**
  * A class which contains utilities to make OAuth2 specific calls.
  */
 public class OAuthUtil {
+
+  public static PrivateKey readPKCS8Pem(String key) throws Exception {
+        key = key.replace("-----BEGIN PRIVATE KEY-----", "");
```
rmstar, line 75: Please fix the indentation. See https://wiki.cask.co/display/CE/Coding+Standard

rmstar (commented):
Can you please add tests?

VictorD-Veolia:
I have proposed a more industrialized implementation of SA Authentication [here](https://github.com/data-integrations/http/pull/68). The main difference with this PR is that I used `com.google.auth` classes (`GoogleCredentials` and `ServiceAccountCredentials`) so that I did not had to implement any parsing/decoding of the SA JSON Key.

bdmogal:
@ropeck thank you for your contribution. Unfortunately, it looks like this capability was already implemented in https://github.com/data-integrations/http/pull/80. Could you please take a look if it suffices, and close this PR?

PR #61, opened 2021-05-28
Fix IncrementAnIndexPaginationIterator

VictorD-Veolia:
Fix "Increment An Index" Pagination to match the expected behavior when "maxIndex" is null ("If empty, pagination will happen until the page with no elements.").

### DIFF
--- a/src/main/java/io/cdap/plugin/http/source/common/pagination/page/JsonPage.java
+++ b/src/main/java/io/cdap/plugin/http/source/common/pagination/page/JsonPage.java
@@ -31,9 +31,6 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 
-/**
- * Returns elements from json one by one by given json path.
- */
 class JsonPage extends BasePage {
   private final String insideElementJsonPathPart;
   private final Iterator<JsonElement> iterator;
@@ -43,9 +40,11 @@ class JsonPage extends BasePage {
   private final BaseHttpSourceConfig config;
   private final List<String> optionalFields;
 
+  // TODO : handle case where the result json object is null
   JsonPage(BaseHttpSourceConfig config, HttpResponse httpResponse) {
     super(httpResponse);
     this.config = config;
+    String body = httpResponse.getBody();
     this.json = JSONUtil.toJsonElement(httpResponse.getBody());
     this.schema = config.getSchema();
     this.optionalFields = getOptionalFields();
@@ -53,7 +52,7 @@ class JsonPage extends BasePage {
     JsonElement jsonElement = json;
     if (json.isJsonObject()) {
       JSONUtil.JsonQueryResponse queryResponse =
-        JSONUtil.getJsonElementByPath(json.getAsJsonObject(), config.getResultPath(), optionalFields);
+              JSONUtil.getJsonElementByPath(json.getAsJsonObject(), config.getResultPath(), optionalFields);
       this.insideElementJsonPathPart = queryResponse.getUnretrievedPath();
       jsonElement = queryResponse.get();
     } else {
@@ -62,11 +61,11 @@ class JsonPage extends BasePage {
 
     if (jsonElement.isJsonArray()) {
       this.iterator = jsonElement.getAsJsonArray().iterator();
-    } else if (jsonElement.isJsonObject()) {
+    } else if (jsonElement.isJsonObject() || jsonElement.isJsonNull()) {
       this.iterator = Collections.singleton(jsonElement).iterator();
     } else {
       throw new IllegalArgumentException(String.format("Element found by '%s' json path is expected to be an object " +
-                                                         "or an array. Primitive found", config.getResultPath()));
+              "or an array. Primitive found", config.getResultPath()));
     }
 
     this.fieldsMapping = config.getFullFieldsMapping();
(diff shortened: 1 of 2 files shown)


bdmogal:
@VictorD-Veolia: possible to share the exact problem that this fixes? Some sample response that this fails for? Also, could you add a test?

VictorD-Veolia:
The "Increment An Index" pagination is not usable at all. Having a  pipeline with an HTTP plugin using "Increment an Index" pagination will throw "Failed to configure pipeline: Stage 'HTTP' encountered : Null error occurred while configuring the stage HTTP." at pipeline startup (with every configuration and independently of input data).

src/main/java/io/cdap/plugin/http/source/common/pagination/page/JsonPage.java
```diff
@@ -31,9 +31,6 @@
 import java.util.List;
 import java.util.Map;
 
-/**
- * Returns elements from json one by one by given json path.
```
rmstar, line 35: why was this removed?

src/main/java/io/cdap/plugin/http/source/common/pagination/page/JsonPage.java
```diff
@@ -43,17 +40,19 @@ class JsonPage extends BasePage {
...
+    String body = httpResponse.getBody();
     this.json = JSONUtil.toJsonElement(httpResponse.getBody());
     this.schema = config.getSchema();
     this.optionalFields = getOptionalFields();
 
     JsonElement jsonElement = json;
     if (json.isJsonObject()) {
       JSONUtil.JsonQueryResponse queryResponse =
-        JSONUtil.getJsonElementByPath(json.getAsJsonObject(), config.getResultPath(), optionalFields);
+              JSONUtil.getJsonElementByPath(json.getAsJsonObject(), config.getResultPath(), optionalFields);
```
rmstar, line 55: revert indent change

src/main/java/io/cdap/plugin/http/source/common/pagination/page/JsonPage.java
```diff
@@ -62,11 +61,11 @@ class JsonPage extends BasePage {
 
     if (jsonElement.isJsonArray()) {
       this.iterator = jsonElement.getAsJsonArray().iterator();
-    } else if (jsonElement.isJsonObject()) {
+    } else if (jsonElement.isJsonObject() || jsonElement.isJsonNull()) {
       this.iterator = Collections.singleton(jsonElement).iterator();
     } else {
       throw new IllegalArgumentException(String.format("Element found by '%s' json path is expected to be an object " +
-                                                         "or an array. Primitive found", config.getResultPath()));
+              "or an array. Primitive found", config.getResultPath()));
```
rmstar, line 68: revert indent change

src/main/java/io/cdap/plugin/http/source/common/pagination/page/JsonPage.java
```diff
@@ -79,56 +78,56 @@ public boolean hasNext() {
 
   /**
    * Converts a next element from json into a json object which is defined by fieldsMapping.
-   *
+   * <p>
    * Example next element:
-   *   {
```
rmstar, line 84: why was this changed?

src/main/java/io/cdap/plugin/http/source/common/pagination/page/JsonPage.java
```diff
@@ -79,56 +78,56 @@ public boolean hasNext() {
...
    * | projectCategory | /fields/project/projectCategory/name      |
    * | isSubtask       | /fields/issuetype/subtask                 |
    * | fixVersions     | /fields/fixVersions                       |
-   *
+   * <p>
    * The result returned by function is:
-   *
+   * <p>
    * {
-   *    "key":"NETTY-13",
```
rmstar, line 122: why was this changed?

src/main/java/io/cdap/plugin/http/source/common/pagination/page/JsonPage.java
```diff
@@ -196,7 +206,7 @@ private List<String> getOptionalFields() {
   public String getPrimitiveByPath(String path) {
     if (json.isJsonObject()) {
       JSONUtil.JsonQueryResponse queryResponse = JSONUtil.getJsonElementByPath(json.getAsJsonObject(),
-                                                                               path, optionalFields);
+              path, optionalFields);
```
rmstar, line 209: revert indent change

src/main/java/io/cdap/plugin/http/source/common/pagination/page/JsonPage.java
```diff
@@ -43,17 +40,19 @@ class JsonPage extends BasePage {
   private final BaseHttpSourceConfig config;
   private final List<String> optionalFields;
 
+  // TODO : handle case where the result json object is null
   JsonPage(BaseHttpSourceConfig config, HttpResponse httpResponse) {
     super(httpResponse);
     this.config = config;
+    String body = httpResponse.getBody();
```
rmstar, line 47: where is this used?

src/main/java/io/cdap/plugin/http/source/common/pagination/page/JsonPage.java
```diff
@@ -43,17 +40,19 @@ class JsonPage extends BasePage {
   private final BaseHttpSourceConfig config;
   private final List<String> optionalFields;
 
+  // TODO : handle case where the result json object is null
```
rmstar, line 43: what happens when result json object is null? are you planning to handle that case in another PR?

src/main/java/io/cdap/plugin/http/source/common/pagination/page/JsonPage.java
```diff
@@ -139,36 +138,47 @@ public boolean hasNext() {
...
-    }
-
-    String jsonString = resultJson.toString();
-    try {
-      StructuredRecord record = StructuredRecordStringConverter.fromJsonString(jsonString, schema);
-      if (numPartiallyRetrieved > 0) {
-        InvalidEntry<StructuredRecord> error =
-          new InvalidEntry<>(1, "Couldn't find all required fields in the record", record);
-        return new PageEntry(error, config.getErrorHandling());
+    if (iterator.hasNext()) {
```
rmstar, line 141: you can avoid one level of indent if you return when there are no entries:
```
if (!iterartor.hasNext()) {
   return new PageEntry(null);
}
...
```

rmstar (commented):
- Please create a Jira that describes the issue and link to it on this PR
- Please add a unit test

PR #64, opened 2021-08-20
fix: Null Errors, using macros in url with pagination

swaptil7:
fix: defer validation at runtime, when url contains a macro avoiding NullPointerExceptions


swaptil7:
@googlebot I signed it!

PR #65, opened 2021-09-02
add the capacity to parse a JSON Object or JSON Array as a string in HTTP source plugins

VictorD-Veolia:
Some APIs might return JSON data with dynamically changing schema for some fields. Today, HTTP Source plugins are not able to handle this kind of data. 

For example, in the following [google API](https://developers.google.com/admin-sdk/directory/reference/rest/v1/users#User) , the field 'customSchema' might have a changing schema.

This pull request add the capacity to parse a JSON object/array to a stringify value.

Here is an example : 

Considering the following input data : 

```
{
	"someField":"someValue",
	"customSchema":{
		"customInnerField1":"innerValue1",
		"customInnerField2":"innerValue2",
		"customInnerField3":"innerValue3"
	},
	"someObject":{
		"someId":"someValue",
		"someType":"someValue"
	}
}
```

Where 'customSchema' might contain a large and unknown number of 'customInnerField'.

With this new feature, the plugin can handle this data if the configuration field 'Authorize Parsing of Objects to String' is set to True, and the field 'customSchema' is defined as STRING in the output schema.

The output record would then be : 

```
{
	"someField":"someValue",
	"customSchema":"{\"customInnerField1\":\"innerValue1\",\"customInnerField2\":\"innerValue2\",\"customInnerField3\":\"innerValue3\"}",
	"someObject":{
		"someId":"someValue",
		"someType":"someValue"
	}
}
```

### DIFF
--- a/docs/HTTP-batchsource.md
+++ b/docs/HTTP-batchsource.md
@@ -203,6 +203,10 @@ can be omitted as long as the field is present in schema.
 
 **CSV Skip First Row:** Whether to skip the first row of the HTTP response. This is usually set if the first row is a header row.
 
+**Authorize Parsing of Objects to String:** Whether to allow the parsing of JSON Objects/Arrays to string.
+If set to true, every JSON Objects/Arrays in the record will be parsed as a string field if defined as string in the output schema.
+This option can be used to handle JSON that contain fields with dynamically changing schema
+
 ### Basic Authentication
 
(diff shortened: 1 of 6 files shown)


rmstar (commented):
- Please create a Jira describing the feature and link to it here
 - Please fix merge conflicts
 - Please add a unit test

PR #67, opened 2021-09-13
Add 'client_credentials' OAuth2 authentication flow

VictorD-Veolia:
Currently, HTTP plugins handle only the "refresh_token" flow for OAuth2 Authentication.
This PR adds support for "[client_credentials](https://datatracker.ietf.org/doc/html/rfc6749#section-4.4)" flow as well.


avirmani:
@VictorD-Veolia - could you please add unit tests here as well.

VictorD-Veolia:
Sorry, we don't plan to add unit test on this PR. 
There is no tests on the existing OAuth2 authentication feature. Implementing unit test on this new OAuth2 flow would take more time than we can afford.

PR #69, opened 2021-09-24
Fix Wait Time Between Pages feature

VictorD-Veolia:
The "Wait Time Between Pages" configuration for HTTP sources was visible in configuration widget only when pagination was disabled.

It is now visible when pagination is enabled.


seanzhougoogle (approved):

rmstar (approved):
please squash commits

PR #75, opened 2022-01-21
Fix "Increment an Index" pagination for HTTP Source plugin

VictorD-Veolia:
In the HTTP source plugin, if the URL contains a macro, and the pagination type is set as "Increment an Index", the plugin's configuration validation fails with a Null Pointer Exception.

This PR fixes this bug by cancelling URL validation when URL contains macro.


avirmani:
@VictorD-Veolia - could you add a unit test for this as well, pls.

PR #79, opened 2022-02-11
[PLUGIN-949] missing key value error msg

sumitnigamcs:
missing key value error msg

### DIFF
--- a/src/main/java/io/cdap/plugin/http/source/batch/HttpBatchSourceConfig.java
+++ b/src/main/java/io/cdap/plugin/http/source/batch/HttpBatchSourceConfig.java
@@ -24,4 +24,113 @@ public class HttpBatchSourceConfig extends BaseHttpSourceConfig {
   protected HttpBatchSourceConfig(String referenceName) {
     super(referenceName);
   }
+
+  private HttpBatchSourceConfig(HttpBatchSourceConfigBuilder builder) {
+    super(builder.referenceName);
+    this.url = builder.url;
+    this.httpMethod = builder.httpMethod;
+    this.headers = builder.headers;
+    this.format = builder.format;
+    this.oauth2Enabled = builder.oauth2Enabled;
+    this.errorHandling = builder.errorHandling;
+    this.retryPolicy = builder.retryPolicy;
+    this.maxRetryDuration = builder.maxRetryDuration;
+    this.connectTimeout = builder.connectTimeout;
+    this.readTimeout = builder.readTimeout;
+    this.paginationType = builder.paginationType;
+    this.verifyHttps = builder.verifyHttps;
+  }
+
+  public static HttpBatchSourceConfigBuilder builder() {
+    return new HttpBatchSourceConfigBuilder();
+  }
+
+  /**
+   * Builder for HttpBatchSourceConfig
+   */
+  public static class HttpBatchSourceConfigBuilder {
+
+    private String referenceName;
+    private String url;
+    private String httpMethod;
+    private String headers;
+    private String format;
+    private String oauth2Enabled;
+    private String errorHandling;
+    private String retryPolicy;
+    private Long maxRetryDuration;
+    private Integer connectTimeout;
+    private Integer readTimeout;
+    private String paginationType;
+    private String verifyHttps;
+
+    public HttpBatchSourceConfigBuilder setReferenceName (String referenceName) {
+      this.referenceName = referenceName;
+      return this;
+    }
+
+    public HttpBatchSourceConfigBuilder setUrl(String url) {
+      this.url = url;
+      return this;
+    }
+
+    public HttpBatchSourceConfigBuilder setHttpMethod(String httpMethod) {
+      this.httpMethod = httpMethod;
+      return this;
+    }
+
+    public HttpBatchSourceConfigBuilder setHeaders(String headers) {
+      this.headers = headers;
+      return this;
+    }
+
+    public HttpBatchSourceConfigBuilder setFormat(String format) {
+      this.format = format;
+      return this;
+    }
+
+    public HttpBatchSourceConfigBuilder setOauth2Enabled(String oauth2Enabled) {
+      this.oauth2Enabled = oauth2Enabled;
+      return this;
+    }
+
+    public HttpBatchSourceConfigBuilder setErrorHandling(String errorHandling) {
+      this.errorHandling = errorHandling;
+      return this;
+    }
+
+    public HttpBatchSourceConfigBuilder setRetryPolicy(String retryPolicy) {
+      this.retryPolicy = retryPolicy;
+      return this;
+    }
+
+    public HttpBatchSourceConfigBuilder setMaxRetryDuration(Long maxRetryDuration) {
+      this.maxRetryDuration = maxRetryDuration;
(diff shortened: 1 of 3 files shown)


sumitnigamcs:
@rmstar Please review for PLUGIN-949

src/main/java/io/cdap/plugin/http/source/common/BaseHttpSourceConfig.java
```diff
@@ -38,9 +39,11 @@
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
```
rmstar: looks like these imports are not used. please remove unused imports

src/main/java/io/cdap/plugin/http/source/common/BaseHttpSourceConfig.java
```diff
@@ -870,6 +873,9 @@ public static Long toLong(String value, String propertyName) {
     String[] mappings = keyValueString.split(",");
     for (String map : mappings) {
       String[] columns = map.split(":");
+      if (columns.length < 2) {
```
rmstar: Is this error seen during configure time? If yes, you should add the failure to the failure collector

rmstar (commented):
Can you please look into adding a unit test?

src/main/java/io/cdap/plugin/http/source/common/BaseHttpSourceConfig.java
sumitnigamcs: removed

sumitnigamcs:
> Can you please look into adding a unit test?

Done @rmstar , Please review.
CC @itsankit-google

src/main/java/io/cdap/plugin/http/source/batch/HttpBatchSourceConfig.java
```diff
@@ -24,4 +24,23 @@
   protected HttpBatchSourceConfig(String referenceName) {
     super(referenceName);
   }
+
+  public HttpBatchSourceConfig(String referenceName, String url, String httpMethod, String headers,
```
rmstar: This can be package private. Also annotate with @VisibleForTesting

src/test/java/io/cdap/plugin/http/source/common/HttpBatchSourceConfigTest.java
```diff
@@ -0,0 +1,21 @@
+package io.cdap.plugin.http.source.common;
+
+import io.cdap.plugin.http.source.batch.HttpBatchSourceConfig;
+import org.junit.Test;
+
+/**
+ * Unit tests for HttpBatchSourceConfig
+ */
+public class HttpBatchSourceConfigTest {
+    private static final String MOCK_STAGE = "mockStage";
```
rmstar: remove unused

src/test/java/io/cdap/plugin/http/source/common/HttpBatchSourceConfigTest.java
```diff
@@ -0,0 +1,21 @@
...
+
+    private static final HttpBatchSourceConfig CONFIG = new HttpBatchSourceConfig("test",
```
rmstar: make this a local variable in the test

src/test/java/io/cdap/plugin/http/source/common/HttpBatchSourceConfigTest.java
```diff
@@ -0,0 +1,21 @@
...
+            "http://localhost", "GET", "Auth:", "JSON", "false",
+            "", "", 600L, 120, 120,
+            "NONE", "true");
+
+    @Test (expected = IllegalArgumentException.class)
+    public void testValidConfig() {
```
rmstar: Please rename to something like testMissingKeyValue

src/main/java/io/cdap/plugin/http/source/common/BaseHttpSourceConfig.java
sumitnigamcs: created https://cdap.atlassian.net/browse/PLUGIN-1109 for the same

src/test/java/io/cdap/plugin/http/source/common/HttpBatchSourceConfigTest.java
sumitnigamcs: done

src/test/java/io/cdap/plugin/http/source/common/HttpBatchSourceConfigTest.java
sumitnigamcs: done

src/test/java/io/cdap/plugin/http/source/common/HttpBatchSourceConfigTest.java
sumitnigamcs: done

src/main/java/io/cdap/plugin/http/source/batch/HttpBatchSourceConfig.java
sumitnigamcs: added annotation.
changing access specifier, sorry I didn't get this because its being called from test package

itsankit-google: I would rather prefer to have a public builder and private constructor, which is the practice we follow. @rmstar WDYT?

rmstar: 👍 yes we can have a builder

sumitnigamcs: @rmstar Please confirm

rmstar: Please follow @itsankit-google 's suggestion

sumitnigamcs: Added the same now.

src/main/java/io/cdap/plugin/http/source/batch/HttpBatchSourceConfig.java
```diff
@@ -24,4 +24,109 @@
...
+    this.connectTimeout = builder.connectTimeout;
+    this.readTimeout = builder.readTimeout;
+    this.paginationType = builder.paginationType;
+    this.verifyHttps = builder.verifyHttps;
+  }
+
+  /**
+   * Builder for HttpBatchSourceConfig
+   */
+  public static class HttpBatchSourceConfigBuilder {
```
itsankit-google, line 51: Add a public static builder method in the config class to return new object of builder.
See this for reference, https://github.com/data-integrations/google-cloud/blob/6fa17791c6dfd10ec26a37564c4c618401206ee1/src/main/java/io/cdap/plugin/gcp/bigquery/sink/BigQuerySinkConfig.java#L680

sumitnigamcs: done

rmstar (approved):

PR #80, opened 2022-03-25
[PLUGIN-692] Add Service Account auth to http

sumengwang:
# [PLUGIN-692] Add Service Account auth to http

## Description
Added Service Account auth to http batch source. Currently the unit tests won't pass

## Links
[PLUGIN-692](https://cdap.atlassian.net/browse/PLUGIN-692)

## Screenshots
image

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -80,6 +80,7 @@
     <commons.version>3.9</commons.version>
     <common.codec.version>1.12</common.codec.version>
     <gson.version>2.8.5</gson.version>
+    <googleapis.version>1.32.1</googleapis.version>
     <hadoop.version>2.3.0</hadoop.version>
     <httpcomponents.version>4.5.9</httpcomponents.version>
     <hydrator.version>2.4.0</hydrator.version>
@@ -323,6 +324,11 @@
       <groupId>org.apache.httpcomponents</groupId>
       <artifactId>httpclient</artifactId>
       <version>${httpcomponents.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.api-client</groupId>
+      <artifactId>google-api-client</artifactId>
+      <version>${googleapis.version}</version>
     </dependency>
     <dependency>
       <groupId>com.google.code.gson</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -80,7 +80,9 @@
     <commons.version>3.9</commons.version>
     <common.codec.version>1.12</common.codec.version>
     <gson.version>2.8.5</gson.version>
-    <googleapis.version>1.32.1</googleapis.version>
+    <googleauth.version>1.6.0</googleauth.version>
+    <guava.version>31.1-jre</guava.version>
+    <jwt.version>3.19.0</jwt.version>
     <hadoop.version>2.3.0</hadoop.version>
     <httpcomponents.version>4.5.9</httpcomponents.version>
     <hydrator.version>2.4.0</hydrator.version>
@@ -326,9 +328,19 @@
       <version>${httpcomponents.version}</version>
     </dependency>
     <dependency>
-      <groupId>com.google.api-client</groupId>
-      <artifactId>google-api-client</artifactId>
-      <version>${googleapis.version}</version>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+      <version>${guava.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.auth</groupId>
+      <artifactId>google-auth-library-oauth2-http</artifactId>
+      <version>${googleauth.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>com.auth0</groupId>
+      <artifactId>java-jwt</artifactId>
+      <version>${jwt.version}</version>
     </dependency>
     <dependency>
       <groupId>com.google.code.gson</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -81,7 +81,7 @@
     <common.codec.version>1.12</common.codec.version>
     <gson.version>2.8.5</gson.version>
     <googleauth.version>1.6.0</googleauth.version>
-    <jwt.version>3.19.0</jwt.version>
+    <guava.version>31.1-jre</guava.version>
     <hadoop.version>2.3.0</hadoop.version>
     <httpcomponents.version>4.5.9</httpcomponents.version>
     <hydrator.version>2.4.0</hydrator.version>
@@ -332,9 +332,9 @@
       <version>${googleauth.version}</version>
     </dependency>
     <dependency>
-      <groupId>com.auth0</groupId>
-      <artifactId>java-jwt</artifactId>
-      <version>${jwt.version}</version>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+      <version>${guava.version}</version>
     </dependency>
     <dependency>
       <groupId>com.google.code.gson</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -76,12 +76,11 @@
 
   <properties>
     <awaitility.version>3.1.6</awaitility.version>
-    <cdap.version>6.1.1</cdap.version>
+    <cdap.version>6.7.0-SNAPSHOT</cdap.version>
     <commons.version>3.9</commons.version>
     <common.codec.version>1.12</common.codec.version>
     <gson.version>2.8.5</gson.version>
     <googleauth.version>1.6.0</googleauth.version>
-    <guava.version>31.1-jre</guava.version>
     <jwt.version>3.19.0</jwt.version>
     <hadoop.version>2.3.0</hadoop.version>
     <httpcomponents.version>4.5.9</httpcomponents.version>
@@ -326,11 +325,6 @@
       <groupId>org.apache.httpcomponents</groupId>
       <artifactId>httpclient</artifactId>
       <version>${httpcomponents.version}</version>
-    </dependency>
-    <dependency>
-      <groupId>com.google.guava</groupId>
-      <artifactId>guava</artifactId>
-      <version>${guava.version}</version>
     </dependency>
     <dependency>
       <groupId>com.google.auth</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -82,7 +82,7 @@
     <gson.version>2.8.5</gson.version>
     <hadoop.version>2.3.0</hadoop.version>
     <httpcomponents.version>4.5.9</httpcomponents.version>
-    <hydrator.version>2.4.0-SNAPSHOT</hydrator.version>
+    <hydrator.version>2.4.0</hydrator.version>
     <jackson.version>2.9.9</jackson.version>
     <junit.version>4.11</junit.version>
     <jython.version>2.7.1</jython.version>
--- a/pom.xml
+++ b/pom.xml
@@ -76,15 +76,14 @@
 
   <properties>
     <awaitility.version>3.1.6</awaitility.version>
-    <cdap.version>6.7.0-SNAPSHOT</cdap.version>
+    <cdap.version>6.8.0-SNAPSHOT</cdap.version>
     <commons.version>3.9</commons.version>
     <common.codec.version>1.12</common.codec.version>
     <gson.version>2.8.5</gson.version>
-    <googleauth.version>1.6.0</googleauth.version>
-    <guava.version>31.1-jre</guava.version>
+    <googleauth.version>0.1.0</googleauth.version>
     <hadoop.version>2.3.0</hadoop.version>
     <httpcomponents.version>4.5.9</httpcomponents.version>
-    <hydrator.version>2.4.0</hydrator.version>
+    <hydrator.version>2.10.0-SNAPSHOT</hydrator.version>
     <jackson.version>2.9.9</jackson.version>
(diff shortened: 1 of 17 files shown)


src/main/java/io/cdap/plugin/http/source/common/BaseHttpSourceConfig.java
```diff
@@ -276,9 +283,32 @@
   @Macro
   protected Long waitTimeBetweenPages;
 
-  @Name(PROPERTY_OAUTH2_ENABLED)
-  @Description("If true, plugin will perform OAuth2 authentication.")
-  protected String oauth2Enabled;
```
tivv, line 281: Was it released already? If yes, do we provide backwards compatibility?

sumengwang: Oauth2 was already released. I believe I only changed ways to determine if Oauth2 is enabled, the actual workflow stays the same

tivv: If it's already released, I don't think we can take the property back. We risk not being able to read old configuration, correct me if I am wrong

sumengwang: ok, it makes sense.

sumengwang: But one of the issue I have is if I leave it there, the bottom of the page will show this extra element:
image

src/main/java/io/cdap/plugin/http/source/batch/HttpBatchSourceConfig.java
```diff
@@ -63,12 +64,18 @@ public static HttpBatchSourceConfigBuilder builder() {
...
     private String paginationType;
     private String verifyHttps;
+    private String authType;
 
     public HttpBatchSourceConfigBuilder setReferenceName (String referenceName) {
       this.referenceName = referenceName;
       return this;
     }
 
+    public HttpBatchSourceConfigBuilder setOauth2Enabled(String oauth2Enabled) {
```
tivv: nit: any specific reason to move this around?

src/main/java/io/cdap/plugin/http/source/common/BaseHttpSourceConfig.java
```diff
@@ -784,14 +856,30 @@ public void validate() {
...
-    // Validate OAuth2 properties
-    if (!containsMacro(PROPERTY_OAUTH2_ENABLED) && this.getOauth2Enabled()) {
-      String reasonOauth2 = "OAuth2 is enabled";
-      assertIsSet(getAuthUrl(), PROPERTY_AUTH_URL, reasonOauth2);
-      assertIsSet(getTokenUrl(), PROPERTY_TOKEN_URL, reasonOauth2);
-      assertIsSet(getClientId(), PROPERTY_CLIENT_ID, reasonOauth2);
-      assertIsSet(getClientSecret(), PROPERTY_CLIENT_SECRET, reasonOauth2);
-      assertIsSet(getRefreshToken(), PROPERTY_REFRESH_TOKEN, reasonOauth2);
+    // Validate Authentication properties
+    switch (this.authType) {
```
tivv: You need to honor PROPERTY_OAUTH2_ENABLED to ensure old configs still work as before

src/main/java/io/cdap/plugin/http/source/common/BaseHttpSourceConfig.java
```diff
@@ -784,14 +856,30 @@ public void validate() {
...
+      case "oAuth2":
+        String reasonOauth2 = "OAuth2 is enabled";
+        assertIsSet(getAuthUrl(), PROPERTY_AUTH_URL, reasonOauth2);
+        assertIsSet(getTokenUrl(), PROPERTY_TOKEN_URL, reasonOauth2);
+        assertIsSet(getClientId(), PROPERTY_CLIENT_ID, reasonOauth2);
+        assertIsSet(getClientSecret(), PROPERTY_CLIENT_SECRET, reasonOauth2);
+        assertIsSet(getRefreshToken(), PROPERTY_REFRESH_TOKEN, reasonOauth2);
+        break;
+      case "serviceAccount":
```
tivv: Please use constants for this values

src/main/java/io/cdap/plugin/http/source/common/exceptions/InvalidPropertyTypeException.java
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright © 2020 Cask Data, Inc.
```
tivv: Use proper year

sumengwang: I copied this from the google drive plugin. Do I just update the year?

tivv: yep

src/main/java/io/cdap/plugin/http/source/common/http/HttpClient.java
```diff
@@ -126,15 +126,17 @@ private CloseableHttpClient createHttpClient() throws IOException {
 
     ArrayList<Header> clientHeaders = new ArrayList<>();
 
-    // oAuth2
-    if (config.getOauth2Enabled()) {
```
tivv, line 130: You must honor this property for backwards compatibility

src/main/java/io/cdap/plugin/http/source/common/http/HttpClient.java
```diff
@@ -126,8 +126,16 @@ private CloseableHttpClient createHttpClient() throws IOException {
...
+    // oAuth2 - old version
+    if (config.getOauth2Enabled()) {
+      String accessToken = OAuthUtil.getAccessTokenByRefreshToken(HttpClients.createDefault(), config.getTokenUrl(),
+                                                                  config.getClientId(), config.getClientSecret(),
+                                                                  config.getRefreshToken());
+      clientHeaders.add(new BasicHeader("Authorization", "Bearer " + accessToken));
+    }
 
+    // auth check
+    AuthType authType = config.getAuthType();
```
tivv: Can you just do something like `if (config.getOauth2Enabled()) { authType = OAUTH2; }`? This would reduce code duplication and ensure we don't add the header twice if for whatever reason both configs are set

src/main/java/io/cdap/plugin/http/source/common/http/AuthType.java
```diff
@@ -0,0 +1,61 @@
...
+
+import io.cdap.plugin.http.source.common.BaseHttpSourceConfig;
+import io.cdap.plugin.http.source.common.exceptions.InvalidPropertyTypeException;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Enum for data range presenting.
```
tivv: I am a bit confused by this comment. Could you please rephrase?

src/main/java/io/cdap/plugin/http/source/common/http/AuthType.java
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  /**
+   * Returns the AuthType.
+   *
+   * @param value the value is string type.
+   * @return The AuthType
+   */
+  public static AuthType fromValue(String value) {
+    return Arrays.stream(AuthType.values()).filter(authtype -> authtype.getValue().equals(value))
```
tivv, line 52: Why not simply `valueOf`?

src/main/java/io/cdap/plugin/http/source/common/http/OAuthUtil.java
```diff
@@ -54,5 +67,52 @@ public static String getAccessTokenByRefreshToken(CloseableHttpClient httpclient
     JsonElement jsonElement = JSONUtil.toJsonObject(responseString).get("access_token");
     return jsonElement.getAsString();
   }
+
+  public static String getAccessTokenByServiceAccount(BaseHttpSourceConfig config) throws IOException {
+    GoogleCredentials credential;
+    String jwtToken = "";
+    if (config.isServiceAccountJson()) {
+      InputStream jsonInputStream = new ByteArrayInputStream(config.getServiceAccountJson().getBytes());
```
tivv: Ensure you pass UTF-8 otherwise it's current encoding-dependent. Also it's a good habit to put any InputStreams into try with resources block to ensure it's closed.

src/main/java/io/cdap/plugin/http/source/common/http/OAuthUtil.java
```diff
@@ -54,5 +67,52 @@ public static String getAccessTokenByRefreshToken(CloseableHttpClient httpclient
...
+      credential = GoogleCredentials.fromStream(jsonInputStream);
+      jwtToken = generateJwt(5000, credential);
```
tivv: How did you decide on 5 seconds? Also it's better to have it at least as a constant as it makes code more readable

sumengwang: This is 5000 seconds based on the documentation. I'm not sure what time range I should choose

src/main/java/io/cdap/plugin/http/source/common/http/OAuthUtil.java
sumengwang: The function itself throws IOException so I did not add the try block

src/main/java/io/cdap/plugin/http/source/common/http/AuthType.java
sumengwang: I copied this file from the google drive plugin. I didn't really look at it into details

src/main/java/io/cdap/plugin/http/source/common/http/OAuthUtil.java
tivv: Oh, I see, you convert it to millis. 5000 seconds is probably fine, but please make a constant that clearly says it's seconds in it's name

fernst: Also 5000 seconds is an odd time. I would expect it to be aligned to an hour (3600 seconds) or a day (86400 seconds).

I also think this value should be in a configuration option for the user to configure. Some users might not feel comfortable with long living JWT tokens.

pom.xml
```diff
@@ -80,9 +80,12 @@
     <commons.version>3.9</commons.version>
     <common.codec.version>1.12</common.codec.version>
     <gson.version>2.8.5</gson.version>
+    <googleauth.version>1.6.0</googleauth.version>
+    <guava.version>31.1-jre</guava.version>
```
tivv: I'd try to use guava version from hydrator. Otherwise strange things may happen

tivv: And BTW, where are you using it? I don't see guava references in the source

src/main/java/io/cdap/plugin/http/source/common/http/AuthType.java
tivv: Please use valueOf, this looks strange

src/main/java/io/cdap/plugin/http/source/common/http/OAuthUtil.java
```diff
@@ -54,5 +68,54 @@ public static String getAccessTokenByRefreshToken(CloseableHttpClient httpclient
     JsonElement jsonElement = JSONUtil.toJsonObject(responseString).get("access_token");
     return jsonElement.getAsString();
   }
+
+  public static String getAccessTokenByServiceAccount(BaseHttpSourceConfig config) throws IOException {
+    GoogleCredentials credential;
+    Integer expiryLength = (config.getJwtTokenExpiryLength() != null) ? config.getJwtTokenExpiryLength() : 3600;
```
tivv: I think this var a can be `int`

sumengwang: The reason I use Integer is because it can be compared with null. I can't do that with int

pom.xml
sumengwang: The GoogleCredentials is asking for one of the package from guava. It won't let me build before I include it.

src/main/java/io/cdap/plugin/http/source/common/http/OAuthUtil.java
tivv: I understand for config, but I don't think you need to do it for `expiryLength` var

pom.xml
tivv: Hm. Why don't they include it themselves? Usually it's up to an artifact to include any dependency needed

src/main/java/io/cdap/plugin/http/source/common/http/AuthType.java
tivv: Discussed offline

src/main/java/io/cdap/plugin/http/source/common/http/OAuthUtil.java
tivv: Not try-catch, but try-with-resources

tivv (approved):

sumengwang:
Question: how should I handle the scope for SA auth?

PR #83, opened 2022-05-03
[PLUGIN-826] skip some validations when url contains macro

sumengwang:
# [PLUGIN-826] skip some validations when url contains macro

## Description
Currently when the url contains a macro that will get evaluated at runtime, validation will throw error since the url will get a null value. Hence skipping the validations that required the url when the url contains macro.

[PLUGIN-826](https://cdap.atlassian.net/browse/PLUGIN-826)

### DIFF
--- a/src/main/java/io/cdap/plugin/http/source/common/BaseHttpSourceConfig.java
+++ b/src/main/java/io/cdap/plugin/http/source/common/BaseHttpSourceConfig.java
@@ -768,31 +768,31 @@ public abstract class BaseHttpSourceConfig extends ReferencePluginConfig {
       try {
         // replace with placeholder with anything just during pagination
         new URI(getUrl().replaceAll(PAGINATION_INDEX_PLACEHOLDER_REGEX, "0"));
+
+        // Validate HTTP Error Handling Map
+        if (!containsMacro(PROPERTY_HTTP_ERROR_HANDLING)) {
+          List<HttpErrorHandlerEntity> httpErrorsHandlingEntries = getHttpErrorHandlingEntries();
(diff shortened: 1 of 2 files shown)


src/main/java/io/cdap/plugin/http/source/common/BaseHttpSourceConfig.java
```diff
@@ -768,98 +768,97 @@ public void validate(FailureCollector failureCollector) {
       try {
         // replace with placeholder with anything just during pagination
         new URI(getUrl().replaceAll(PAGINATION_INDEX_PLACEHOLDER_REGEX, "0"));
```
seanzhougoogle, line 770: you will still get NPE if url is null ,right ?

src/main/java/io/cdap/plugin/http/source/common/BaseHttpSourceConfig.java
```diff
@@ -833,7 +833,7 @@ public void validate(FailureCollector failureCollector) {
                                         propertiesShouldBeNull.remove(PROPERTY_INDEX_INCREMENT));
           propertiesShouldBeNull.remove(PROPERTY_MAX_INDEX); // can be both null and non null
 
-          if (!url.contains(PAGINATION_INDEX_PLACEHOLDER)) {
+          if (url != null && !url.contains(PAGINATION_INDEX_PLACEHOLDER)) {
```
seanzhougoogle: !containsMacro(PROPERTY_URL)

seanzhougoogle (approved):
one minor comment else LGTM

PR #85, opened 2022-10-04
CDAP-19657: add fqn support

mrahanjam:
This PR adds FQN to the Salesforce batch plugin code.

Jira: [CDAP-19657](https://cdap.atlassian.net/browse/CDAP-19657)


src/main/java/io/cdap/plugin/http/sink/batch/HTTPSinkConfig.java
```diff
@@ -271,6 +274,12 @@ public Map<String, String> getRequestHeadersMap() {
     return convertHeadersToMap(requestHeaders);
   }
 
+  public String getReferenceNameOrNormalizedFQN() {
+    return Strings.isNullOrEmpty(referenceName)
```
fernst: Nit: maintain this in one line

fernst (approved):

PR #89, opened 2022-11-03
[CDAP-20012] Adding support for missing keys in http plugin and adding some error …

Aryan-Verma:
Bug: https://cdap.atlassian.net/browse/CDAP-20012

The customer was facing the issue that if there is a missing non-nullable field in the HTTP plugin, they get an error stating `No matching schema found`. Instead they wanted the missing fields to get interpreted as null in the output schema fields.

The root cause of the issue is: The input fields of the plugin are converted to JsonObject. If there is a missing field, a partial json path to the field is retrieved. The proposed change instead returns a null object for that missing field.

Also added some error logs for suggesting user to set the `nullable` option in the case when there is a missing field in the input schema and nullable is not set.


MrRahulSharma (approved):
LGTM

PR #96, opened 2022-12-07
[PLUGIN-1447] Creating the buffered reader only when iterating over the pages.

MrRahulSharma:
Jira link : https://cdap.atlassian.net/browse/PLUGIN-1447
The BufferedReader is created lazily, to make sure that an opened stream is used to iterate the page.

Validations:
1. Along with custom pagination logic the page iteration works as expected.
2. In case of Text page without pagination the page iteration works as expected.


itsankit-google (approved):
Please add a unit test.

Otherwise, LGTM.

PR #106, opened 2023-02-08
Added null check when we get an empty http response

Nikitapaliwal123:
Added null check when we get an empty http response

### DIFF
--- a/src/main/java/io/cdap/plugin/http/source/common/http/HttpResponse.java
+++ b/src/main/java/io/cdap/plugin/http/source/common/http/HttpResponse.java
@@ -80,6 +80,9 @@ public class HttpResponse implements Closeable {
   public byte[] getBytes() {
     if (bytes == null) {
       HttpEntity responseEntity = response.getEntity();
+      if(responseEntity == null) {
+        return bytes;
+      }
       try {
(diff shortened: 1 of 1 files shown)


src/main/java/io/cdap/plugin/http/source/common/http/HttpResponse.java
```diff
@@ -80,6 +80,9 @@ public String getBody() {
   public byte[] getBytes() {
     if (bytes == null) {
       HttpEntity responseEntity = response.getEntity();
+      if(responseEntity == null) {
+        return bytes;
```
MrRahulSharma: Return a zero length array instead.

src/main/java/io/cdap/plugin/http/source/common/http/HttpResponse.java
```diff
@@ -80,6 +80,10 @@ public String getBody() {
   public byte[] getBytes() {
     if (bytes == null) {
       HttpEntity responseEntity = response.getEntity();
+      if(responseEntity == null) {
```
MrRahulSharma, line 83: you can use 'return new byte[0]' as well.

MrRahulSharma (approved):
LGTM

PR #107, opened 2023-02-08
Added null check in case when access token is not present.

Nikitapaliwal123:
Added null check in case when access token is not present.

### DIFF
--- a/src/main/java/io/cdap/plugin/http/source/common/http/OAuthUtil.java
+++ b/src/main/java/io/cdap/plugin/http/source/common/http/OAuthUtil.java
@@ -62,7 +62,7 @@ public class OAuthUtil {
 
     JsonElement jsonElement = JSONUtil.toJsonObject(responseString).get("access_token");
     if (jsonElement == null) {
-         throw new IOException("Access token not found");
+         throw new IllegalArgumentException("Access token not found");
     }
     return jsonElement.getAsString();
   }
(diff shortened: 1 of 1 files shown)


src/main/java/io/cdap/plugin/http/source/common/http/OAuthUtil.java
```diff
@@ -60,6 +61,9 @@ public static String getAccessTokenByRefreshToken(CloseableHttpClient httpclient
     String responseString = EntityUtils.toString(response.getEntity(), "UTF-8");
 
     JsonElement jsonElement = JSONUtil.toJsonObject(responseString).get("access_token");
+    if (jsonElement == null) {
+         throw new IOException("Access token not found");
```
MrRahulSharma: This should not be an IOException.

src/main/java/io/cdap/plugin/http/source/common/http/OAuthUtil.java
```diff
@@ -21,6 +21,7 @@
 import com.google.gson.JsonElement;
 import io.cdap.plugin.http.source.common.BaseHttpSourceConfig;
 import io.cdap.plugin.http.source.common.pagination.page.JSONUtil;
+import org.apache.commons.httpclient.auth.InvalidCredentialsException;
```
MrRahulSharma: Remove this line.

MrRahulSharma (approved):

MrRahulSharma:
Please squash the changes into a single commit and then merge.

PR #109, opened 2023-02-17
PLUGIN-1515 fix http sink when batch size > 1

albertshau:
Fix the HTTP sink to properly write messages in the OutputFormat instead of in the transform and destroy methods of the sink, since the destroy method never gets called in Spark. This ensures that messages are flushed when the RecordWriter is closed.

Also fixed it to properly handle unknown input schemas instead of throwing a NullPointerException, and random checkstyle and unit test failures

### DIFF
--- a/src/main/java/io/cdap/plugin/http/sink/batch/HTTPRecordWriter.java
+++ b/src/main/java/io/cdap/plugin/http/sink/batch/HTTPRecordWriter.java
@@ -1,0 +1,230 @@
+/*
+ * Copyright © 2023 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.plugin.http.sink.batch;
+
+import io.cdap.cdap.api.data.format.StructuredRecord;
+import io.cdap.cdap.api.data.schema.Schema;
+import io.cdap.cdap.format.StructuredRecordStringConverter;
+import org.apache.hadoop.mapreduce.RecordWriter;
+import org.apache.hadoop.mapreduce.TaskAttemptContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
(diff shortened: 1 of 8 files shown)


src/main/java/io/cdap/plugin/http/sink/batch/HTTPRecordWriter.java
```diff
@@ -0,0 +1,230 @@
...
+      } else if (config.getMessageFormat().equals("Form")) {
+        message = createFormMessage(input);
+        contentType = " application/x-www-form-urlencoded";
+      } else if (config.getMessageFormat().equals("Custom")) {
+        message = createCustomMessage(config.getBody(), input);
+        contentType = " text/plain";
+      }
+      messages.append(message).append(config.getDelimiterForMessages());
+    }
+    StringTokenizer tokens = new StringTokenizer(messages.toString().trim(), config.getDelimiterForMessages());
```
rmstar, line 81: Might be more efficient to keep track of the number of messages in the current batch instead of tokenizing `messages` on every write() call

src/main/java/io/cdap/plugin/http/sink/batch/HTTPRecordWriter.java
```diff
@@ -0,0 +1,230 @@
...
+        throw new IllegalStateException("Error opening url connection. Reason: " + e.getMessage(), e);
+      } catch (IOException e) {
+        LOG.warn("Error making {} request to url {} with headers {}.", config.getMethod(), config.getMethod(), headers);
+        exception = e;
+      } finally {
+        if (conn != null) {
+          conn.disconnect();
+        }
+      }
+      retries++;
```
rmstar, line 153: should we backoff between retries?

src/main/java/io/cdap/plugin/http/sink/batch/HTTPRecordWriter.java
```diff
@@ -0,0 +1,230 @@
...
+  }
+
+  @Override
+  public void close(TaskAttemptContext taskAttemptContext) throws IOException, InterruptedException {
+    // Process remaining messages after batch executions.
+    if (!messages.toString().isEmpty()) {
+      try {
+        executeHTTPService();
+      } catch (Exception e) {
+        throw new RuntimeException("Error while executing http request for remaining input messages " +
```
rmstar, line 94: Is there a reason why we want to wrap the IOException thrown by executeHTTPService() in a RuntimeException?
Also, if we want to throw RuntimeException, we might want to set the cause as `e`.

rmstar (approved):
some minor comments which I believe are actually in the existing code that was moved to the record writer

albertshau:
everything you mentioned would be good improvements (as well as other things), but wanted to keep this PR to only refactoring (in addition to fixing things that causes the build to break)

rmstar:
>

rmstar:
sg, lgtm 👍

PR #110, opened 2023-02-21
PLUGIN-1515 fix http sink when batch size > 1

albertshau:
Fix the HTTP sink to properly write messages in the OutputFormat instead of in the transform and destroy methods of the sink, since the destroy method never gets called in Spark. This ensures that messages are flushed when the RecordWriter is closed.

Also fixed it to properly handle unknown input schemas instead of throwing a NullPointerException

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -21,7 +21,7 @@
   <name>HTTP Plugins</name>
   <groupId>io.cdap</groupId>
   <artifactId>http-plugins</artifactId>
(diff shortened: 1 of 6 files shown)


albertshau:
cherry-pick of #109 , with a few changes to make it build, like removing snapshot from dependencies.

snehalakshmisha (approved):

PR #113, opened 2023-04-05
Changed the logic to use HttpHost constructor for creating auth scope

Nikitapaliwal123:
Changed the logic to use HttpHost constructor for creating auth scope in place of its create method.
JIRA : https://cdap.atlassian.net/browse/PLUGIN-1577


MrRahulSharma (approved):
LGTM

PR #114, opened 2023-04-11
[PLUGIN-1577] Changed the logic to use HttpHost constructor for creating AuthScope

MrRahulSharma:
Changed the logic to use HttpHost constructor for creating auth scope in place of its create method.
Cherry-pick : #113
JIRA : https://cdap.atlassian.net/browse/PLUGIN-1577

### DIFF
--- a/src/main/java/io/cdap/plugin/http/source/common/http/HttpClient.java
+++ b/src/main/java/io/cdap/plugin/http/source/common/http/HttpClient.java
@@ -107,7 +107,8 @@ public class HttpClient implements Closeable {
     // basic auth
     CredentialsProvider credentialsProvider = new BasicCredentialsProvider();
     if (!Strings.isNullOrEmpty(config.getUsername()) && !Strings.isNullOrEmpty(config.getPassword())) {
-      AuthScope authScope = new AuthScope(HttpHost.create(config.getUrl()));
+      URI uri = URI.create(config.getUrl());
+      AuthScope authScope = new AuthScope(new HttpHost(uri.getHost(), uri.getPort(), uri.getScheme()));
       credentialsProvider.setCredentials(authScope,
                                          new UsernamePasswordCredentials(config.getUsername(), config.getPassword()));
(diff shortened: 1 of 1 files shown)


sumitjnn:
Please add the Jira id to PR title

src/main/java/io/cdap/plugin/http/source/common/http/HttpClient.java
```diff
@@ -107,7 +107,8 @@ private CloseableHttpClient createHttpClient() throws IOException {
     // basic auth
     CredentialsProvider credentialsProvider = new BasicCredentialsProvider();
     if (!Strings.isNullOrEmpty(config.getUsername()) && !Strings.isNullOrEmpty(config.getPassword())) {
-      AuthScope authScope = new AuthScope(HttpHost.create(config.getUrl()));
+      URI uri = URI.create(config.getUrl());
+      AuthScope authScope = new AuthScope(new HttpHost(uri.getHost(), uri.getPort(), uri.getScheme()));
```
sumitjnn, line 111: Nit: Should we use `HttpHost.create(URI uri)` which seems to be doing the same thing internally?

sumitjnn (commented):
Please check if we can add a unit test

sumitjnn (approved):

src/main/java/io/cdap/plugin/http/source/common/http/HttpClient.java
MrRahulSharma: I believe there might be a mixup with org.apache.commons.httpclient.HttpHost and org.apache.http.HttpHost(used here) class. org.apache.commons.httpclient.HttpHost have the method with URI as argument but org.apache.http.HttpHost does not.

MrRahulSharma:
> Please check if we can add a unit test

Sure, let me create a follow up item post this change to investigate the addition of UT for this change. Not blocking the cherry-pick on this comment.

PR #120, opened 2023-04-25
[PLUGIN-1592] Fixed bug causing auto-detect for service accounts to fail

dli357:


sumengwang (approved):
Thank you for the quick fix

PR #135, opened 2023-06-16
[PLUGIN-1635] Auth Token recreation in case of token expiry 

MrRahulSharma:
This to ensure that a new set of Auth headers are created in case of Auth Token Expiry.
JIRA: https://cdap.atlassian.net/browse/PLUGIN-1635

### DIFF
--- a/src/main/java/io/cdap/plugin/http/common/BaseHttpConfig.java
+++ b/src/main/java/io/cdap/plugin/http/common/BaseHttpConfig.java
@@ -16,6 +16,7 @@
 
 package io.cdap.plugin.http.common;
 
+import com.google.auth.oauth2.AccessToken;
 import com.google.common.base.Strings;
 import io.cdap.cdap.api.annotation.Description;
 import io.cdap.cdap.api.annotation.Macro;
@@ -25,11 +26,7 @@ import io.cdap.cdap.etl.api.validation.InvalidConfigPropertyException;
 import io.cdap.plugin.common.ReferencePluginConfig;
 import io.cdap.plugin.http.common.http.AuthType;
 import io.cdap.plugin.http.common.http.OAuthUtil;
-import org.apache.http.Header;
-import org.apache.http.impl.client.HttpClients;
-import org.apache.http.message.BasicHeader;
 import java.io.File;
-import java.io.IOException;
 import java.util.Optional;
 import javax.annotation.Nullable;
 
@@ -337,7 +334,7 @@ public abstract class BaseHttpConfig extends ReferencePluginConfig {
                 boolean propertiesAreValid = validateServiceAccount(failureCollector);
                 if (propertiesAreValid) {
                     try {
-                        String  accessToken = OAuthUtil.getAccessTokenByServiceAccount(this);
+                        AccessToken accessToken = OAuthUtil.getAccessToken(this);
                     } catch (Exception e) {
                         failureCollector.addFailure("Unable to authenticate given service account info. ",
                                         "Please make sure all infomation entered correctly")
@@ -355,30 +352,6 @@ public abstract class BaseHttpConfig extends ReferencePluginConfig {
                 }
                 break;
         }
-    }
-
-    public Header getAuthorizationHeader() throws IOException {
-
-        // auth check
-        AuthType authType = getAuthType();
-
-        // backward compatibility
-        if (getOauth2Enabled()) {
-            authType = AuthType.OAUTH2;
-        }
-
-        switch (authType) {
-            case OAUTH2:
-                String accessToken = OAuthUtil.getAccessTokenByRefreshToken(HttpClients.createDefault(), getTokenUrl(),
-                        getClientId(), getClientSecret(),
-                        getRefreshToken());
-                return new BasicHeader("Authorization", "Bearer " + accessToken);
-            case SERVICE_ACCOUNT:
-                // get accessToken from service account
-                accessToken = OAuthUtil.getAccessTokenByServiceAccount(this);
-                return new BasicHeader("Authorization", "Bearer " + accessToken);
-        }
-        return null;
     }
 
     public static void assertIsSet(Object propertyValue, String propertyName, String reason) {
--- a/src/main/java/io/cdap/plugin/http/common/http/OAuthUtil.java
+++ b/src/main/java/io/cdap/plugin/http/common/http/OAuthUtil.java
@@ -15,6 +15,7 @@
  */
 package io.cdap.plugin.http.common.http;
 
+import com.google.auth.oauth2.AccessToken;
 import com.google.auth.oauth2.GoogleCredentials;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableSet;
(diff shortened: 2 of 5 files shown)


src/main/java/io/cdap/plugin/http/common/pagination/BaseHttpPaginationIterator.java
```diff
@@ -84,13 +84,20 @@ protected boolean visitPageAndCheckStatusCode() throws IOException {
     if (response != null) { // close previous response
       response.close();
     }
+    if (this.shouldRetry()) {
+      getHttpClient().close(); // In case of retry attempt, close the previous HttpClient instance
```
sau42shri: Could this give a null pointer exception, in case the API isn't able to create an http client

src/main/java/io/cdap/plugin/http/common/pagination/BaseHttpPaginationIterator.java
```diff
@@ -84,13 +84,20 @@ protected boolean visitPageAndCheckStatusCode() throws IOException {
...
     currentPageUrl = nextPageUrl;
     httpStatusCode = response.getStatusCode();
-    RetryableErrorHandling errorHandlingStrategy = httpErrorHandler.getErrorHandlingStrategy(httpStatusCode);
 
-    return !errorHandlingStrategy.shouldRetry();
+    return !shouldRetry();
+  }
+
+  private boolean shouldRetry() {
+    RetryableErrorHandling errorHandlingStrategy = httpErrorHandler.getErrorHandlingStrategy(httpStatusCode);
```
sau42shri: As a general coding practice, may be a good idea to check the object being null, to avoid any null handling exception.

src/main/java/io/cdap/plugin/http/common/pagination/BaseHttpPaginationIterator.java
MrRahulSharma: getHttpClient returns the httpClient instance, which is created in the constructor and is never set to null. Hence I believe this can not be null at this point.

src/main/java/io/cdap/plugin/http/common/pagination/BaseHttpPaginationIterator.java
MrRahulSharma: The httpErrorHandler is created in the constructor and not set to null any where in the code. Hence it should not be null at this point.

Aryan-Verma (approved):

src/main/java/io/cdap/plugin/http/common/pagination/BaseHttpPaginationIterator.java
```diff
@@ -83,11 +83,13 @@ public BaseHttpPaginationIterator(BaseHttpSourceConfig config, PaginationIterato
   protected boolean visitPageAndCheckStatusCode() throws IOException {
     if (response != null) { // close previous response
       response.close();
+      getHttpClient().close();
```
sumitjnn: Creation of new http client for every request seems to be inefficient, is it possible to just set new auth token in header with existing http client?

MrRahulSharma: HttpClient does not provide, any API to update the AuthenticationHeader. But let me check if there is any other way to reuse the HttpClient object created here.

src/main/java/io/cdap/plugin/http/common/BaseHttpConfig.java
```diff
@@ -157,6 +159,12 @@ public abstract class BaseHttpConfig extends ReferencePluginConfig {
     @Macro
     protected String serviceAccountScope;
 
+    /**
+     * The credential object to help with getting the access token
+     */
+    @Nullable
+    private OAuth2Credentials credentials;
```
albertshau: the config class defines configuration properties that the user can supply, it is not meant for putting arbitrary variables of different types.

src/main/java/io/cdap/plugin/http/common/BaseHttpConfig.java
```diff
@@ -369,16 +400,13 @@ public Header getAuthorizationHeader() throws IOException {
 
         switch (authType) {
             case OAUTH2:
-                String accessToken = OAuthUtil.getAccessTokenByRefreshToken(HttpClients.createDefault(), getTokenUrl(),
-                        getClientId(), getClientSecret(),
-                        getRefreshToken());
-                return new BasicHeader("Authorization", "Bearer " + accessToken);
+                this.credentials = OAuthUtil.getUserCredentials(this);
```
albertshau: it's not good to make the config stateful like this, it makes it hard to reason about the state and easy to introduce bugs. It's better to have this return the credential instead of setting a value.

src/main/java/io/cdap/plugin/http/common/http/OAuthUtil.java
```diff
@@ -16,86 +16,66 @@
...
-    CloseableHttpResponse response = httpclient.execute(httppost);
-    String responseString = EntityUtils.toString(response.getEntity(), "UTF-8");
 
-    JsonElement jsonElement = JSONUtil.toJsonObject(responseString).get("access_token");
-    if (jsonElement == null) {
-         throw new IllegalArgumentException("Access token not found");
-    }
-    return jsonElement.getAsString();
+  public static OAuth2Credentials getUserCredentials(BaseHttpConfig config) {
+    UserCredentials userCredentials = new UserCredentials(config.getClientId(),
```
albertshau: I think this makes certain assumptions specific to google oauth apis. For example, it looks like it throws an exception if expires_in does not exist. However, not all providers return that.

Salesforce, for example, returns when the token was issued, but not when it expires (https://developer.salesforce.com/docs/atlas.en-us.196.0.api_rest.meta/api_rest/intro_understanding_refresh_token_oauth.htm). We can't use this, otherwise pipelines that were working before can start failing depending on their particular oauth response

MrRahulSharma: Good catch Albert.

src/main/java/io/cdap/plugin/http/common/BaseHttpConfig.java
```diff
@@ -367,29 +360,11 @@ public void validate(FailureCollector failureCollector) {
...
-                    return new BasicHeader(key, requestMetadata.get(key).get(0));
-                }
-            }
-        }
-
-        return null;
-    }
-
-    private void createOAuthCredentials() throws IOException {
+    public AccessToken getAccessToken() throws IOException {
```
albertshau: method should be annotated as `@Nullable`

src/main/java/io/cdap/plugin/http/sink/batch/HTTPRecordWriter.java
```diff
@@ -110,8 +110,10 @@ private void executeHTTPService() throws IOException {
 
       Map<String, String> headers = config.getRequestHeadersMap();
 
-      Header authorizationHeader = config.getAuthorizationHeader();
-      if (authorizationHeader != null) {
+      AccessToken accessToken = config.getAccessToken();
```
albertshau: existing issue, but this record writer fetches a new access token for each request. This is especially bad if the batch size is 1, then it is getting a new token for every single record. It should instead reuse the token unless it's about to expire.

MrRahulSharma: Added

albertshau (approved):

sumitjnn (approved):

PR #142, opened 2023-09-01
[PLUGIN-1676] Added Json batch support in HTTP sink

psainics:
### Bug 
Previously when selecting batch size `> 1` JSON messages were joined with `\n`
The end service failed to parse JSON messages with `\n`
### Update
JSON messages should be joined as defined in the JSON format `[ {} , {} ]`

For services that accept object instead of an array as input we let user define a `Json Batch Key` (Optional) that is used to wrap the JSON array

## UI Field
image

## UI Field (Filter)
The UI field is only visible when `messageFormat == JSON` and `batchSize > 1`

## UI Field (Validation)
The value for key has allowed characters (a-z, A-Z, 0-9, _, -, .)

## Batch Size (Validation)
Previously there was no validation on the batch size , causing illegal user inputs to be valid.
New validation let's user only enter a valid +ve number greater than 0 

## Unit Test
6 Unit test cases were
### Test for batch size
- testHTTPSinkWith`NegativeBatchSize`
- testHTTPSinkWith`ZeroBatchSize`
- testHTTPSinkWith`PositiveBatchSize`

### Test for valid batch key value
- testHTTPSinkWithBatchKeyContaining`Space`
- testHTTPSinkWithBatchKeyContaining`DoubleQuotes`
- testHTTPSinkWithBatchKeyContaining`AlphaNumeric`

image

## Docs 
Updated docs for the new field added to UI

## Code change
- Messages are stored in a buffer array , when the array length is equal to batch-size we call the function `flushMessageBuffer` that moves all messages from the array to a single string that will be set as the message body.
 The function is responsible for joining all the message as per format selected by user
- Long if - else if - else condition are changes to switch case
- Update createFormMessage to url encode only the key and value
- Added fieldToString to keep all messages (form/json/custom) have same result for each field

## Jira Link
 - [Jira Link](https://cdap.atlassian.net/browse/PLUGIN-1676)

### DIFF
--- a/docs/HTTP-batchsink.md
+++ b/docs/HTTP-batchsink.md
@@ -14,6 +14,20 @@ Properties
 
 **batchSize:** Batch size. Defaults to 1. (Macro enabled)
 
+**Write JSON As Array:** Whether to write the JSON as an array. Defaults to false. (Macro enabled)
+
+When set to true, the payload will be written as an array of JSON objects.
+Example - If batch size is 2, then the payload will be `[{"key":"val"}, {"key":"val"}]`
+
+When false, the payload will be JSON objects separated by a delimiter.
+Example - If batch size is 2, delimiter is "\n" , then the payload will be `{"key":"val"}\n{"key":"val"}`
+
+**Json Batch Key** Optional key to be used for wrapping json array as object
+Leave empty for no wrapping of the array. Ignored if Write JSON As Array is set to false. (Macro Enabled)
+
+Example - If batch size is 2 and json batch key is "data", then the payload will be
+`{"data": [{"key":"val"}, {"key":"val"}]}` instead of `[{"key":"val"}, {"key":"val"}]`
+
 **messageFormat:** Format to send messsage in. Options are JSON, Form, Custom. Defaults to JSON. (Macro enabled)
 
 **body:** Optional custom message. This is required if the message format is set to 'Custom'.
--- a/src/main/java/io/cdap/plugin/http/common/http/MessageFormatType.java
+++ b/src/main/java/io/cdap/plugin/http/common/http/MessageFormatType.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright © 2023 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.cdap.plugin.http.common.http;
+
+import io.cdap.plugin.http.common.EnumWithValue;
+
+/**
+ * An enum which represent a type of message format of an HTTP request body.
+ */
+public enum MessageFormatType implements EnumWithValue {
+
+  JSON,
+  FORM,
+  CUSTOM;
+
+  @Override
+  public String getValue() {
+    return name();
+  }
+
+  @Override
+  public String toString() {
+    return this.getValue();
+  }
+}
--- a/src/main/java/io/cdap/plugin/http/sink/batch/HTTPRecordWriter.java
+++ b/src/main/java/io/cdap/plugin/http/sink/batch/HTTPRecordWriter.java
@@ -17,11 +17,11 @@
 package io.cdap.plugin.http.sink.batch;
 
 import com.google.auth.oauth2.AccessToken;
+
 import io.cdap.cdap.api.data.format.StructuredRecord;
 import io.cdap.cdap.api.data.schema.Schema;
-import io.cdap.cdap.format.StructuredRecordStringConverter;
-
 import io.cdap.plugin.http.common.http.OAuthUtil;
+
 import org.apache.hadoop.mapreduce.RecordWriter;
 import org.apache.hadoop.mapreduce.TaskAttemptContext;
 import org.apache.http.Header;
@@ -31,20 +31,15 @@ import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.io.OutputStream;
-import java.io.UnsupportedEncodingException;
 import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
 import java.net.ProtocolException;
 import java.net.URL;
-import java.net.URLEncoder;
 import java.security.KeyManagementException;
 import java.security.NoSuchAlgorithmException;
 import java.security.cert.X509Certificate;
-import java.util.HashMap;
 import java.util.Map;
-import java.util.StringTokenizer;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
+
 import javax.net.ssl.HostnameVerifier;
 import javax.net.ssl.HttpsURLConnection;
 import javax.net.ssl.SSLContext;
@@ -57,52 +52,34 @@ import javax.net.ssl.X509TrustManager;
  */
 public class HTTPRecordWriter extends RecordWriter<StructuredRecord, StructuredRecord> {
   private static final Logger LOG = LoggerFactory.getLogger(HTTPRecordWriter.class);
-  private static final String REGEX_HASHED_VAR = "#s*(\\w+)";
-
   private final HTTPSinkConfig config;
-  private StringBuilder messages = new StringBuilder();
+  private final MessageBuffer messageBuffer;
   private String contentType;
-
   private AccessToken accessToken;
 
-  HTTPRecordWriter(HTTPSinkConfig config) {
+  HTTPRecordWriter(HTTPSinkConfig config, Schema inputSchema) {
     this.config = config;
     this.accessToken = null;
+    this.messageBuffer = new MessageBuffer(
+            config.getMessageFormat(), config.getJsonBatchKey(), config.shouldWriteJsonAsArray(),
+            config.getDelimiterForMessages(), config.getCharset(), config.getBody(), inputSchema
+    );
   }
 
   @Override
   public void write(StructuredRecord input, StructuredRecord unused) throws IOException {
-    String message = null;
     if (config.getMethod().equals("POST") || config.getMethod().equals("PUT")) {
-      if (config.getMessageFormat().equals("JSON")) {
-        message = StructuredRecordStringConverter.toJsonString(input);
-        contentType = "application/json";
-      } else if (config.getMessageFormat().equals("Form")) {
-        message = createFormMessage(input);
-        contentType = " application/x-www-form-urlencoded";
-      } else if (config.getMessageFormat().equals("Custom")) {
-        message = createCustomMessage(config.getBody(), input);
-        contentType = " text/plain";
-      }
-      messages.append(message).append(config.getDelimiterForMessages());
+      messageBuffer.add(input);
     }
-    StringTokenizer tokens = new StringTokenizer(messages.toString().trim(), config.getDelimiterForMessages());
-    if (config.getBatchSize() == 1 || tokens.countTokens() == config.getBatchSize()) {
-      executeHTTPService();
+    if (config.getBatchSize() == messageBuffer.size()) {
+      flushMessageBuffer();
     }
   }
 
   @Override
-  public void close(TaskAttemptContext taskAttemptContext) throws IOException, InterruptedException {
+  public void close(TaskAttemptContext taskAttemptContext) throws IOException {
     // Process remaining messages after batch executions.
-    if (!messages.toString().isEmpty()) {
-      try {
-        executeHTTPService();
-      } catch (Exception e) {
-        throw new RuntimeException("Error while executing http request for remaining input messages " +
-                                     "after the batch execution. " + e);
-      }
-    }
+    flushMessageBuffer();
   }
 
   private void executeHTTPService() throws IOException {
@@ -110,6 +87,7 @@ public class HTTPRecordWriter extends RecordWriter<StructuredRecord, StructuredR
     int retries = 0;
     IOException exception = null;
     do {
+      exception = null;
       HttpURLConnection conn = null;
 
       Map<String, String> headers = config.getRequestHeadersMap();
@@ -148,14 +126,14 @@ public class HTTPRecordWriter extends RecordWriter<StructuredRecord, StructuredR
             conn.addRequestProperty("Content-Type", contentType);
           }
         }
-        if (messages.length() > 0) {
+        if (!messageBuffer.isEmpty()) {
           conn.setDoOutput(true);
           try (OutputStream outputStream = conn.getOutputStream()) {
-            outputStream.write(messages.toString().trim().getBytes(config.getCharset()));
+            outputStream.write(messageBuffer.getMessage().trim().getBytes(config.getCharset()));
           }
         }
         responseCode = conn.getResponseCode();
-        messages.setLength(0);
+        messageBuffer.clear();
         if (config.getFailOnNon200Response() && !(responseCode >= 200 && responseCode < 300)) {
           exception = new IOException("Received error response. Response code: " + responseCode);
         }
@@ -175,48 +153,6 @@ public class HTTPRecordWriter extends RecordWriter<StructuredRecord, StructuredR
     if (exception != null) {
       throw exception;
     }
-  }
-
-  private String createFormMessage(StructuredRecord input) {
-    boolean first = true;
-    String formMessage = null;
-    StringBuilder sb = new StringBuilder("");
-    for (Schema.Field field : input.getSchema().getFields()) {
-      if (first) {
-        first = false;
-      } else {
-        sb.append("&");
-      }
-      sb.append(field.getName());
-      sb.append("=");
-      sb.append((String) input.get(field.getName()));
-    }
-    try {
-      formMessage = URLEncoder.encode(sb.toString(), config.getCharset());
-    } catch (UnsupportedEncodingException e) {
-      throw new IllegalStateException("Error encoding Form message. Reason: " + e.getMessage(), e);
-    }
-    return formMessage;
-  }
-
-  private String createCustomMessage(String body, StructuredRecord input) {
-    String customMessage = body;
-    Matcher matcher = Pattern.compile(REGEX_HASHED_VAR).matcher(customMessage);
-    HashMap<String, String> findReplaceMap = new HashMap();
-    while (matcher.find()) {
-      if (input.get(matcher.group(1)) != null) {
-        findReplaceMap.put(matcher.group(1), (String) input.get(matcher.group(1)));
-      } else {
-        throw new IllegalArgumentException(String.format(
-          "Field %s doesnt exist in the input schema.", matcher.group(1)));
-      }
-    }
-    Matcher replaceMatcher = Pattern.compile(REGEX_HASHED_VAR).matcher(customMessage);
-    while (replaceMatcher.find()) {
-      String val = replaceMatcher.group().replace("#", "");
-      customMessage = (customMessage.replace(replaceMatcher.group(), findReplaceMap.get(val)));
-    }
-    return customMessage;
   }
 
   private void disableSSLValidation() {
@@ -247,4 +183,14 @@ public class HTTPRecordWriter extends RecordWriter<StructuredRecord, StructuredR
     };
     HttpsURLConnection.setDefaultHostnameVerifier(allHostsValid);
   }
+
+  private void flushMessageBuffer() throws IOException {
+    if (messageBuffer.isEmpty()) {
+      return;
+    }
+    contentType = messageBuffer.getContentType();
+    executeHTTPService();
+    messageBuffer.clear();
+  }
+
 }
--- a/src/main/java/io/cdap/plugin/http/sink/batch/HTTPSinkConfig.java
+++ b/src/main/java/io/cdap/plugin/http/sink/batch/HTTPSinkConfig.java
@@ -27,13 +27,14 @@ import io.cdap.cdap.etl.api.FailureCollector;
 import io.cdap.plugin.common.ReferenceNames;
 
 import io.cdap.plugin.http.common.BaseHttpConfig;
+import io.cdap.plugin.http.common.http.MessageFormatType;
+
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
-import java.util.Optional;
 import java.util.Set;
 import javax.annotation.Nullable;
 import javax.ws.rs.HttpMethod;
@@ -45,6 +46,8 @@ public class HTTPSinkConfig extends BaseHttpConfig {
   public static final String URL = "url";
   public static final String METHOD = "method";
   public static final String BATCH_SIZE = "batchSize";
+  public static final String WRITE_JSON_AS_ARRAY = "writeJsonAsArray";
+  public static final String JSON_BATCH_KEY = "jsonBatchKey";
   public static final String DELIMETER_FOR_MESSAGE = "delimiterForMessages";
   public static final String MESSAGE_FORMAT = "messageFormat";
   public static final String BODY = "body";
@@ -75,6 +78,19 @@ public class HTTPSinkConfig extends BaseHttpConfig {
   @Description("Batch size. Defaults to 1. (Macro Enabled)")
   @Macro
   private final Integer batchSize;
+
+  @Name(WRITE_JSON_AS_ARRAY)
+  @Nullable
+  @Description("Whether to write json as array. Defaults to false. (Macro Enabled)")
+  @Macro
+  private final Boolean writeJsonAsArray;
+
+  @Name(JSON_BATCH_KEY)
+  @Nullable
+  @Description("Optional key to be used for wrapping json array as object. " +
+          "Leave empty for no wrapping of the array (Macro Enabled)")
+  @Macro
+  private final String jsonBatchKey;
 
   @Name(DELIMETER_FOR_MESSAGE)
   @Nullable
@@ -150,7 +166,8 @@ public class HTTPSinkConfig extends BaseHttpConfig {
                         @Nullable String requestHeaders, String charset,
                         boolean followRedirects, boolean disableSSLValidation, @Nullable int numRetries,
                         @Nullable int readTimeout, @Nullable int connectTimeout, boolean failOnNon200Response,
-                        String oauth2Enabled, String authType) {
+                        String oauth2Enabled, String authType, @Nullable String jsonBatchKey,
+                        Boolean writeJsonAsArray) {
     super(referenceName);
     this.url = url;
     this.method = method;
@@ -166,6 +183,8 @@ public class HTTPSinkConfig extends BaseHttpConfig {
     this.readTimeout = readTimeout;
     this.connectTimeout = connectTimeout;
     this.failOnNon200Response = failOnNon200Response;
+    this.jsonBatchKey = jsonBatchKey;
+    this.writeJsonAsArray = writeJsonAsArray;
     this.oauth2Enabled = oauth2Enabled;
     this.authType = authType;
   }
@@ -186,6 +205,8 @@ public class HTTPSinkConfig extends BaseHttpConfig {
     connectTimeout = builder.connectTimeout;
     readTimeout = builder.readTimeout;
     failOnNon200Response = builder.failOnNon200Response;
+    jsonBatchKey = builder.jsonBatchKey;
+    writeJsonAsArray = builder.writeJsonAsArray;
     oauth2Enabled = builder.oauth2Enabled;
     authType = builder.authType;
   }
@@ -201,7 +222,7 @@ public class HTTPSinkConfig extends BaseHttpConfig {
     builder.method = copy.getMethod();
     builder.batchSize = copy.getBatchSize();
     builder.delimiterForMessages = copy.getDelimiterForMessages();
-    builder.messageFormat = copy.getMessageFormat();
+    builder.messageFormat = copy.getMessageFormat().getValue();
     builder.body = copy.getBody();
     builder.requestHeaders = copy.getRequestHeaders();
     builder.charset = copy.getCharset();
@@ -228,13 +249,20 @@ public class HTTPSinkConfig extends BaseHttpConfig {
     return batchSize;
   }
 
-  @Nullable
-  public String getDelimiterForMessages() {
-    return delimiterForMessages;
+  public boolean shouldWriteJsonAsArray() {
+    return writeJsonAsArray != null && writeJsonAsArray;
   }
 
-  public String getMessageFormat() {
-    return messageFormat;
+  public String getJsonBatchKey() {
+    return jsonBatchKey;
+  }
+
+  public String getDelimiterForMessages() {
+    return Strings.isNullOrEmpty(delimiterForMessages) ? "\n" : delimiterForMessages;
+  }
+
+  public MessageFormatType getMessageFormat() {
+    return MessageFormatType.valueOf(messageFormat.toUpperCase());
   }
 
   @Nullable
@@ -319,6 +347,11 @@ public class HTTPSinkConfig extends BaseHttpConfig {
         .withConfigProperty(METHOD);
     }
 
+    if (!containsMacro(BATCH_SIZE) && batchSize != null && batchSize < 1) {
+         collector.addFailure("Batch size must be greater than 0.", null)
+                 .withConfigProperty(BATCH_SIZE);
+    }
+
     if (!containsMacro(NUM_RETRIES) && numRetries < 0) {
       collector.addFailure("Number of Retries cannot be a negative number.", null)
         .withConfigProperty(NUM_RETRIES);
@@ -370,6 +403,8 @@ public class HTTPSinkConfig extends BaseHttpConfig {
     private String url;
     private String method;
     private Integer batchSize;
+    private Boolean writeJsonAsArray;
+    private String jsonBatchKey;
     private String delimiterForMessages;
     private String messageFormat;
     private String body;
@@ -404,6 +439,16 @@ public class HTTPSinkConfig extends BaseHttpConfig {
 
     public Builder setBatchSize(Integer batchSize) {
       this.batchSize = batchSize;
+      return this;
+    }
+
+    public Builder setWriteJsonAsArray(Boolean writeJsonAsArray) {
+      this.writeJsonAsArray = writeJsonAsArray;
+      return this;
+    }
+
+    public Builder setJsonBatchKey(String jsonBatchKey) {
+      this.jsonBatchKey = jsonBatchKey;
       return this;
     }
 
--- a/src/main/java/io/cdap/plugin/http/sink/batch/MessageBuffer.java
+++ b/src/main/java/io/cdap/plugin/http/sink/batch/MessageBuffer.java
@@ -1,0 +1,232 @@
+/*
+ * Copyright © 2023 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.plugin.http.sink.batch;
+
+import com.google.common.base.Strings;
+import io.cdap.cdap.api.data.format.StructuredRecord;
+import io.cdap.cdap.api.data.schema.Schema;
+import io.cdap.cdap.format.StructuredRecordStringConverter;
+import io.cdap.plugin.http.common.http.MessageFormatType;
+
+import java.io.IOException;
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.function.Function;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+/**
+ * MessageBuffer is used to store the structured records in a buffer till the batch size is reached.
+ * Once the batch size is reached, the records are converted to the appropriate format and appended to the message.
+ * The message is then returned to the HTTPRecordWriter.
+ */
+public class MessageBuffer {
+  private static final String REGEX_HASHED_VAR = "#s*(\\w+)";
+  private final List<StructuredRecord> buffer;
+  private final String jsonBatchKey;
+  private final Boolean shouldWriteJsonAsArray;
+  private final String delimiterForMessages;
+  private final String charset;
+  private final String customMessageBody;
+  private final Function<List<StructuredRecord>, String> messageFormatter;
+  private final String contentType;
+  private final Schema wrappedMessageSchema;
+
+
+  /**
+   * Constructor for MessageBuffer.
+   *
+   * @param messageFormat          The format of the message. Can be JSON, FORM or CUSTOM.
+   * @param jsonBatchKey           The key to be used for the JSON batch message.
+   * @param shouldWriteJsonAsArray Whether the JSON message should be written as an array.
+   * @param delimiterForMessages   The delimiter to be used for messages.
+   * @param charset                The charset to be used for the message.
+   * @param customMessageBody      The custom message body to be used.
+   */
+  public MessageBuffer(
+          MessageFormatType messageFormat, String jsonBatchKey, boolean shouldWriteJsonAsArray,
+          String delimiterForMessages, String charset, String customMessageBody, Schema inputSchema
+  ) {
+    this.jsonBatchKey = jsonBatchKey;
+    this.delimiterForMessages = delimiterForMessages;
+    this.charset = charset;
+    this.shouldWriteJsonAsArray = shouldWriteJsonAsArray;
+    this.customMessageBody = customMessageBody;
+    this.buffer = new ArrayList<>();
+    switch (messageFormat) {
+      case JSON:
+        messageFormatter = this::formatAsJson;
+        contentType = "application/json";
+        break;
+      case FORM:
+        messageFormatter = this::formatAsForm;
+        contentType = "application/x-www-form-urlencoded";
+        break;
+      case CUSTOM:
+        messageFormatter = this::formatAsCustom;
+        contentType = "text/plain";
+        break;
+      default:
+        throw new IllegalArgumentException("Invalid message format: " + messageFormat);
+    }
+    // A new StructuredRecord is created with the jsonBatchKey as the
+    // field name and the array of records as the value
+    Schema bufferRecordArraySchema = Schema.arrayOf(inputSchema);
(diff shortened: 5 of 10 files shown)


src/main/java/io/cdap/plugin/http/sink/batch/HTTPRecordWriter.java
```diff
@@ -74,28 +78,36 @@ public class HTTPRecordWriter extends RecordWriter<StructuredRecord, StructuredR
...
-      }
-      messages.append(message).append(config.getDelimiterForMessages());
+        switch (config.getMessageFormat()) {
+            case "JSON":
+                message = StructuredRecordStringConverter.toJsonString(input);
+                contentType = "application/json";
+                break;
+            case "Form":
+                message = createFormMessage(input);
+                contentType = "application/x-www-form-urlencoded";
```
albertshau: can't really tell where this is being used (like if it is getting appended to something), but the previous logic had a preceding space (` application...` vs `application...`). Was that wrong to have?

src/main/java/io/cdap/plugin/http/sink/batch/HTTPRecordWriter.java
```diff
@@ -247,4 +256,68 @@ public boolean verify(String hostname, SSLSession session) {
     };
     HttpsURLConnection.setDefaultHostnameVerifier(allHostsValid);
   }
```
albertshau, line 185: style: newline between methods

src/main/java/io/cdap/plugin/http/sink/batch/HTTPRecordWriter.java
```diff
@@ -74,28 +78,36 @@ public class HTTPRecordWriter extends RecordWriter<StructuredRecord, StructuredR
...
-        contentType = "application/json";
-      } else if (config.getMessageFormat().equals("Form")) {
-        message = createFormMessage(input);
-        contentType = " application/x-www-form-urlencoded";
-      } else if (config.getMessageFormat().equals("Custom")) {
-        message = createCustomMessage(config.getBody(), input);
-        contentType = " text/plain";
-      }
-      messages.append(message).append(config.getDelimiterForMessages());
+        switch (config.getMessageFormat()) {
```
albertshau: fix indentation to be consistent (2 spaces)

src/main/java/io/cdap/plugin/http/sink/batch/HTTPRecordWriter.java
```diff
@@ -247,4 +256,68 @@ public boolean verify(String hostname, SSLSession session) {
     };
     HttpsURLConnection.setDefaultHostnameVerifier(allHostsValid);
   }
+  private void flushMessageBuffer() {
+    int batchSize = config.getBatchSize();
+    String jsonBatchKey = config.getJsonBatchKey();
+    boolean useJsonBatchKey = !Strings.isNullOrEmpty(jsonBatchKey) && batchSize > 1;
+      switch (config.getMessageFormat()) {
```
albertshau: indentation

src/main/java/io/cdap/plugin/http/sink/batch/HTTPRecordWriter.java
```diff
@@ -247,4 +256,68 @@ public boolean verify(String hostname, SSLSession session) {
...
+              break;
+          case "Custom":
+              messages.append(String.join(config.getDelimiterForMessages(), messageBuffer));
+              break;
+        default:
+          throw new IllegalArgumentException("Invalid message format.");
+      }
+    messageBuffer.clear();
+  }
+  protected static String fieldToString(StructuredRecord record, Schema.Field field) {
```
albertshau: add a newline between methods

src/main/java/io/cdap/plugin/http/sink/batch/HTTPRecordWriter.java
```diff
@@ -178,25 +191,21 @@ private void executeHTTPService() throws IOException {
...
-    String formMessage = null;
-    StringBuilder sb = new StringBuilder("");
+    StringBuilder sb = new StringBuilder();
     for (Schema.Field field : input.getSchema().getFields()) {
-      if (first) {
-        first = false;
-      } else {
-        sb.append("&");
+      String key = field.getName();
+      String value = fieldToString(input, field);
```
albertshau: this is a change from before that's unrelated to json. It looks like it should be in a different commit and PR. Is there another jira open for it?

docs/HTTP-batchsink.md
```diff
@@ -14,6 +14,9 @@ Properties
 
 **batchSize:** Batch size. Defaults to 1. (Macro enabled)
 
+**Json Batch Key** Optional key to be used for wrapping json array as object " +
```
albertshau: I could not understand what this meant until I saw the code. It would be good to include an example.

src/main/java/io/cdap/plugin/http/sink/batch/HTTPSinkConfig.java
```diff
@@ -406,6 +437,10 @@ public Builder setBatchSize(Integer batchSize) {
       this.batchSize = batchSize;
       return this;
     }
+    public Builder setJsonBatchKey(String jsonBatchKey) {
```
albertshau, line 450: newline between methods

albertshau (changes requested):
this is a backwards incompatible change. Any pipeline that was relying on the previous behavior will break when using this new version.

For these types of fixes, a new optional config property needs to be added. Something like a 'Write Json as Array' option that defaults to the previous behavior (false). Only if somebody sets it to true should the plugin switch to new behavior.

src/main/java/io/cdap/plugin/http/sink/batch/HTTPRecordWriter.java
psainics: When user selects the format `JSON/Form/Custom` we need to set the HTTP header `content-type` accordingly.
```java 
conn.addRequestProperty("Content-Type", contentType);
```
The extra " " is required if this was added manually , but using an HTTP library to add headers it's handled internally.

docs/HTTP-batchsink.md
psainics: Sure 👍 
> Example - If batch size is 2 and json batch key is "data", then the payload will be
`{"data": [{"key":"val"}, {"key":"val"}]}` instead of `[{"key":"val"}, {"key":"val"}]`

src/main/java/io/cdap/plugin/http/sink/batch/HTTPSinkConfig.java
psainics: Updated

src/main/java/io/cdap/plugin/http/sink/batch/HTTPRecordWriter.java
psainics: Updated

src/main/java/io/cdap/plugin/http/sink/batch/HTTPRecordWriter.java
psainics: Updated

src/main/java/io/cdap/plugin/http/sink/batch/HTTPRecordWriter.java
psainics: Updated

src/main/java/io/cdap/plugin/http/sink/batch/HTTPRecordWriter.java
psainics: Updated

src/main/java/io/cdap/plugin/http/sink/batch/HTTPRecordWriter.java
PushpenderSaini0: Changes are reverted

PushpenderSaini0:
## New Ui Field
image

## Docs 
> Write JSON As Array: Whether to write the JSON as an array. Defaults to false. Ignored if batch size is 1. (Macro enabled)

> When set to true, the payload will be written as an array of JSON objects. Example - If batch size is 2, then the payload will be [{"key":"val"}, {"key":"val"}]

> When false, the payload will be JSON objects separated by a delimiter. Example - If batch size is 2, delimiter is “\n” , then the payload will be {"key":"val"}\n{"key":"val"}

PushpenderSaini0:
@albertshau @DJSagarAhire
I have made changes to address the breaking change by adding a new field as suggested , please review the code.

psainics:
## New Ui Field
image

## Docs 
> Write JSON As Array: Whether to write the JSON as an array. Defaults to false. Ignored if batch size is 1. (Macro enabled)

> When set to true, the payload will be written as an array of JSON objects. Example - If batch size is 2, then the payload will be [{"key":"val"}, {"key":"val"}]

> When false, the payload will be JSON objects separated by a delimiter. Example - If batch size is 2, delimiter is “\n” , then the payload will be {"key":"val"}\n{"key":"val"}

src/main/java/io/cdap/plugin/http/sink/batch/HTTPRecordWriter.java
psainics: Changes are reverted

psainics:
@albertshau @DJSagarAhire
I have made changes to address the breaking change by adding a new field as suggested , please review the code.

src/main/java/io/cdap/plugin/http/sink/batch/HTTPRecordWriter.java
```diff
@@ -74,28 +76,36 @@ public class HTTPRecordWriter extends RecordWriter<StructuredRecord, StructuredR
...
+        case "Form":
+          message = createFormMessage(input);
+          contentType = "application/x-www-form-urlencoded";
+          break;
+        case "Custom":
+          message = createCustomMessage(config.getBody(), input);
+          contentType = "text/plain";
+          break;
+        default:
+          throw new IllegalArgumentException("Invalid message format.");
```
albertshau: nit: it's better to include the invalid value in the message `"Invalid message format: " + config.getMessageFormat()`

src/main/java/io/cdap/plugin/http/sink/batch/HTTPSinkConfig.java
```diff
@@ -319,6 +351,22 @@ public void validate(FailureCollector collector) {
...
+    if (!containsMacro(BATCH_SIZE) && batchSize != null && batchSize < 1) {
+         collector.addFailure("Batch size must be greater than 0.", null)
+                 .withConfigProperty(BATCH_SIZE);
+    }
+
+    if (!containsMacro(JSON_BATCH_KEY) && !Strings.isNullOrEmpty(jsonBatchKey)) {
+      Pattern pattern = Pattern.compile("[^a-zA-Z0-9_\\-.]");
+      Matcher matcher = pattern.matcher(jsonBatchKey);
+      if (matcher.find()) {
+        collector.addFailure("JSON Batch Key must be a valid JSON key. " +
```
albertshau: we shouldn't need to do this check, json keys are allowed to have other characters there. We should just use whatever value the user has given.

src/main/java/io/cdap/plugin/http/sink/batch/HTTPRecordWriter.java
```diff
@@ -61,6 +62,7 @@ public class HTTPRecordWriter extends RecordWriter<StructuredRecord, StructuredR
 
   private final HTTPSinkConfig config;
   private StringBuilder messages = new StringBuilder();
+  private ArrayList<String> messageBuffer = new ArrayList<>();
```
albertshau: nit: generally better to declare variables as the more generic interface. In this case, `List<String> messageBuffer = new ArrayList<>()`

src/main/java/io/cdap/plugin/http/sink/batch/HTTPRecordWriter.java
```diff
@@ -247,4 +258,33 @@ public boolean verify(String hostname, SSLSession session) {
     };
     HttpsURLConnection.setDefaultHostnameVerifier(allHostsValid);
   }
+
+  private void flushMessageBuffer() {
+    int batchSize = config.getBatchSize();
+    String jsonBatchKey = config.getJsonBatchKey();
+    boolean useJsonBatchKey = !Strings.isNullOrEmpty(jsonBatchKey) && batchSize > 1;
+    boolean shouldWriteJsonAsArray = config.shouldWriteJsonAsArray();
+    switch (config.getMessageFormat()) {
```
albertshau: having duplicate switch statements here and in the write() method is a code smell. It is cleaner to change the message buffer into a `List<StructuredRecord>` and have the write() method just append to it. Then this flush method is the one that formats the records and performs the http request.

src/main/java/io/cdap/plugin/http/sink/batch/HTTPRecordWriter.java
```diff
@@ -74,28 +76,36 @@ public class HTTPRecordWriter extends RecordWriter<StructuredRecord, StructuredR
...
+          throw new IllegalArgumentException("Invalid message format.");
       }
-      messages.append(message).append(config.getDelimiterForMessages());
+      messageBuffer.add(message);
     }
-    StringTokenizer tokens = new StringTokenizer(messages.toString().trim(), config.getDelimiterForMessages());
-    if (config.getBatchSize() == 1 || tokens.countTokens() == config.getBatchSize()) {
+    int batchSize = config.getBatchSize();
+    if (messageBuffer.size() == batchSize) {
+      flushMessageBuffer();
```
albertshau, line 75: with a method named 'flush', I would expect it to actually write to the http service. It's better to move the executeHTTPService() call into the flush method.

This would also remove a little duplicate code in the close() method.

src/main/java/io/cdap/plugin/http/sink/batch/HTTPRecordWriter.java
```diff
@@ -74,28 +76,36 @@ public class HTTPRecordWriter extends RecordWriter<StructuredRecord, StructuredR
...
       executeHTTPService();
     }
   }
 
   @Override
   public void close(TaskAttemptContext taskAttemptContext) throws IOException, InterruptedException {
     // Process remaining messages after batch executions.
-    if (!messages.toString().isEmpty()) {
+    if (!messageBuffer.isEmpty()) {
+      flushMessageBuffer();
```
albertshau: this is an existing issue a few lines below, but it's not good to catch the Exception from executeHTTPService and rethrow it as a RuntimeException without setting the cause. This results in a bad exception message and loses the stack trace.

It should just call executeHTTPService() without catching, as this method throws IOException already.

src/main/java/io/cdap/plugin/http/sink/batch/HTTPRecordWriter.java
```diff
@@ -247,4 +258,33 @@ public boolean verify(String hostname, SSLSession session) {
...
+
+  private void flushMessageBuffer() {
+    int batchSize = config.getBatchSize();
+    String jsonBatchKey = config.getJsonBatchKey();
+    boolean useJsonBatchKey = !Strings.isNullOrEmpty(jsonBatchKey) && batchSize > 1;
+    boolean shouldWriteJsonAsArray = config.shouldWriteJsonAsArray();
+    switch (config.getMessageFormat()) {
+      case "JSON":
+        String messageBody;
+        if (!shouldWriteJsonAsArray) {
```
albertshau: these checks would be easier to understand if structured as:

```
if (!shouldWriteJsonAsArray) {
  ...
  break;
}
if (useJsonBatchKey) {
  ...
} else {
  ...
}
```

docs/HTTP-batchsink.md
```diff
@@ -14,6 +14,22 @@ Properties
 
 **batchSize:** Batch size. Defaults to 1. (Macro enabled)
 
+**Write JSON As Array:** Whether to write the JSON as an array. Defaults to false. Ignored if batch size is 1.
```
albertshau: It's simpler to have this always apply instead of having it depend on the batch size value. 

It also allows somebody to send arrays of 1 element if they want to for some reason.

docs/HTTP-batchsink.md
```diff
@@ -14,6 +14,22 @@ Properties
...
+(Macro enabled)
+
+When set to true, the payload will be written as an array of JSON objects.
+Example - If batch size is 2, then the payload will be `[{"key":"val"}, {"key":"val"}]`
+
+When false, the payload will be JSON objects separated by a delimiter.
+Example - If batch size is 2, delimiter is "\n" , then the payload will be `{"key":"val"}\n{"key":"val"}`
+
+**Json Batch Key** Optional key to be used for wrapping json array as object " +
+"when batch size is more than 1. Leave empty for no wrapping of the array (Macro Enabled)
```
albertshau: put (Macro Enabled) at the end of the description on the next line.

Similar to above, it's simpler if this setting always applies regardless of the batch size.

src/main/java/io/cdap/plugin/http/sink/batch/HTTPRecordWriter.java
```diff
@@ -247,4 +258,33 @@ public boolean verify(String hostname, SSLSession session) {
...
+          messageBody = String.join(config.getDelimiterForMessages(), messageBuffer);
+        } else if (useJsonBatchKey) {
+          messageBody = String.format("{\"%s\":[%s]}", jsonBatchKey, String.join(",", messageBuffer));
+        } else {
+          messageBody = batchSize == 1 ? messageBuffer.get(0) :
```
albertshau: as mentioned in the doc, I think it's better not to have the format dependent on the batch size.

src/main/java/io/cdap/plugin/http/sink/batch/HTTPRecordWriter.java
```diff
@@ -247,4 +258,33 @@ public boolean verify(String hostname, SSLSession session) {
...
+    String jsonBatchKey = config.getJsonBatchKey();
+    boolean useJsonBatchKey = !Strings.isNullOrEmpty(jsonBatchKey) && batchSize > 1;
+    boolean shouldWriteJsonAsArray = config.shouldWriteJsonAsArray();
+    switch (config.getMessageFormat()) {
+      case "JSON":
+        String messageBody;
+        if (!shouldWriteJsonAsArray) {
+          messageBody = String.join(config.getDelimiterForMessages(), messageBuffer);
+        } else if (useJsonBatchKey) {
+          messageBody = String.format("{\"%s\":[%s]}", jsonBatchKey, String.join(",", messageBuffer));
```
albertshau: Manually manipulating strings like this is error prone. I'm assuming that's why there is that validation check on the characters in the jsonBatchKey, because this doesn't escape the key or anything? Instead, use a json library like the one being used by the StructuredRecordStringConverter. I think something like:

```
private static final JsonStructuredRecordDatumWriter JSON_DATUM_WRITER = new JsonStructuredRecordDatumWriter();
...

private void flushMessages() {
  ...
    StringWriter strWriter = new StringWriter();
    try (JsonWriter writer = new JsonWriter(strWriter)) {
      writer.beginObject().name(jsonBatchKey).beginArray();
      Encoder encoder = new JsonEncoder(writer)
      for (StructuredRecord record : messageBuffer) {
        JSON_DATUM_WRITER.encode(record, encoder);
      }
      writer.endArray().endObject();
    }
    ...
```

albertshau (commented):
there should also be some unit tests for the actual logic around formatting records into a response body. Can refactor the formatting logic into a separate class or method and call it in a new unit test.

src/main/java/io/cdap/plugin/http/sink/batch/HTTPRecordWriter.java
psainics: Updated

src/main/java/io/cdap/plugin/http/sink/batch/HTTPSinkConfig.java
psainics: This check has been removed

src/main/java/io/cdap/plugin/http/sink/batch/HTTPRecordWriter.java
psainics: Updated

src/main/java/io/cdap/plugin/http/sink/batch/HTTPRecordWriter.java
psainics: This logic has been moved to a separate class, duplicate switch was removed

src/main/java/io/cdap/plugin/http/sink/batch/HTTPRecordWriter.java
psainics: Updated, flush method is now responsible for calling the http service

src/main/java/io/cdap/plugin/http/sink/batch/HTTPRecordWriter.java
psainics: Updated
```java
if (!config.shouldWriteJsonAsArray()) {
  // ...
  break;
}
if (useJsonBatchKey) {
 // ...
} else {
  ...
}
```

docs/HTTP-batchsink.md
psainics: Updated to let user use size 1 array

docs/HTTP-batchsink.md
psainics: Updated

src/main/java/io/cdap/plugin/http/sink/batch/HTTPRecordWriter.java
psainics: Updated to let user use size 1 array

src/main/java/io/cdap/plugin/http/sink/batch/HTTPRecordWriter.java
psainics: Due to old gson version in cdap-api, we don't have .jsonValue function.

In new approach i have wrapped the records in a new structure record before calling the .toJsonString

psainics:
Added unit test for the new message buffer class.

> MessageBuffer is used to store the structured records in a buffer till the batch size is reached.
> Once the batch size is reached, the records are converted to the appropriate format and appended to the message.
>The message is then returned to the HTTPRecordWriter.
 
image

docs/HTTP-batchsink.md
```diff
@@ -14,6 +14,20 @@ Properties
...
+**Write JSON As Array:** Whether to write the JSON as an array. Defaults to false. (Macro enabled)
+
+When set to true, the payload will be written as an array of JSON objects.
+Example - If batch size is 2, then the payload will be `[{"key":"val"}, {"key":"val"}]`
+
+When false, the payload will be JSON objects separated by a delimiter.
+Example - If batch size is 2, delimiter is "\n" , then the payload will be `{"key":"val"}\n{"key":"val"}`
+
+**Json Batch Key** Optional key to be used for wrapping json array as object
+Leave empty for no wrapping of the array Ignored if Write JSON As Array is set to false. (Macro Enabled)
```
albertshau: missing period. `... array. Ignored if ...`

src/main/java/io/cdap/plugin/http/sink/batch/HTTPSinkConfig.java
```diff
@@ -76,6 +80,19 @@ public class HTTPSinkConfig extends BaseHttpConfig {
...
 
+  @Name(WRITE_JSON_AS_ARRAY)
+  @Nullable
+  @Description("Whether to write json as array. Defaults to false. (Macro Enabled)")
+  @Macro
+  private final String writeJsonAsArray;
+
+  @Name(JSON_BATCH_KEY)
+  @Nullable
+  @Description("Optional key to be used for wrapping json array as object " +
```
albertshau: add a period. `as object. `

src/main/java/io/cdap/plugin/http/sink/batch/HTTPSinkConfig.java
```diff
@@ -76,6 +80,19 @@ public class HTTPSinkConfig extends BaseHttpConfig {
   @Macro
   private final Integer batchSize;
 
+  @Name(WRITE_JSON_AS_ARRAY)
+  @Nullable
+  @Description("Whether to write json as array. Defaults to false. (Macro Enabled)")
+  @Macro
+  private final String writeJsonAsArray;
```
albertshau: this should just be a Boolean type instead of using a String and parsing it.

src/main/java/io/cdap/plugin/http/sink/batch/MessageBuffer.java
```diff
@@ -0,0 +1,177 @@
...
+  private final List<StructuredRecord> buffer;
+  private final StringBuilder message;
+  private final HTTPSinkConfig config;
+  private static final Map<String, String> contentTypeMap = new HashMap<String, String>() {{
+    put("json", "application/json");
+    put("form", "application/x-www-form-urlencoded");
+    put("custom", "text/plain");
+  }};
+
+  public MessageBuffer(HTTPSinkConfig config) {
```
albertshau: this should just take the arguments that are relevant to the formatting rather than the whole config.

src/main/java/io/cdap/plugin/http/sink/batch/HTTPRecordWriter.java
```diff
@@ -247,4 +180,16 @@ public boolean verify(String hostname, SSLSession session) {
     };
     HttpsURLConnection.setDefaultHostnameVerifier(allHostsValid);
   }
+
+  private void flushMessageBuffer() throws IOException {
+    if (messageBuffer.isEmpty()) {
+      return;
+    }
+    // Stored in messages as at a retry level, we may need to retry the entire batch.
+    messages.append(messageBuffer.getMessage());
```
albertshau: we shouldn't have both the messageBuffer and a messages StringBuilder. It will use extra memory, but more importantly appending here makes the method have side effects. For example, if we later add retries, calling this method multiple times would result in an incorrect request body.

It's better to just remove the StringBuilder and use the messageBuffer in the executeHttpService() method.

src/main/java/io/cdap/plugin/http/sink/batch/MessageBuffer.java
```diff
@@ -0,0 +1,177 @@
...
+  public String getContentType() {
+    return contentTypeMap.getOrDefault(config.getMessageFormat().toLowerCase(), "application/octet-stream");
+  }
+
+  public String getMessage() throws IOException {
+    if (buffer.isEmpty()) {
+      return "";
+    }
+
+    switch (config.getMessageFormat().toLowerCase()) {
```
albertshau: it's better not to have this type of static check (once the class is created, the result of the switch will always be the same) in methods that are repeatedly called.

To avoid this, you can have 3 separate implementations of a `Function<List<StructuredRecord>, String>`, one for json, one for custom, one for form. Then in the constructor of this class, switch on message format to assign the function, and just call it here.

src/main/java/io/cdap/plugin/http/sink/batch/MessageBuffer.java
```diff
@@ -0,0 +1,177 @@
...
+    return buffer.isEmpty();
+  }
+
+  public String getContentType() {
+    return contentTypeMap.getOrDefault(config.getMessageFormat().toLowerCase(), "application/octet-stream");
+  }
+
+  public String getMessage() throws IOException {
+    if (buffer.isEmpty()) {
+      return "";
```
albertshau: I don't think getMessage() is ever currently called with an empty buffer, but an empty json array is more accurately `[]` and not an empty string.

src/main/java/io/cdap/plugin/http/sink/batch/MessageBuffer.java
```diff
@@ -0,0 +1,177 @@
...
+          // field name and the array of records as the value
+          Schema bufferRecordSchema = buffer.get(0).getSchema();
+          Schema bufferRecordArraySchema = Schema.arrayOf(bufferRecordSchema);
+          Schema wrappedMessegeSchema = Schema.recordOf("wrapper",
+                  Schema.Field.of(jsonBatchKey, bufferRecordArraySchema));
+          StructuredRecord wrappedMessageRecord = StructuredRecord.builder(wrappedMessegeSchema)
+                  .set(jsonBatchKey, buffer).build();
+          message.append(StructuredRecordStringConverter.toJsonString(wrappedMessageRecord));
+        } else {
+          message.append(String.format("[%s]", String.join(",", getBufferAsJsonList())));
```
albertshau: this requires one pass to get a List, then another pass to format it into another string. It's more time and space efficient to append it to the buffer in one pass, like what is being done for the 'form' and 'custom' formats, except surrounded by appends for the '[' and ']'

src/main/java/io/cdap/plugin/http/sink/batch/MessageBuffer.java
```diff
@@ -0,0 +1,177 @@
...
+          String jsonBatchKey = config.getJsonBatchKey();
+          // A new StructuredRecord is created with the jsonBatchKey as the
+          // field name and the array of records as the value
+          Schema bufferRecordSchema = buffer.get(0).getSchema();
+          Schema bufferRecordArraySchema = Schema.arrayOf(bufferRecordSchema);
+          Schema wrappedMessegeSchema = Schema.recordOf("wrapper",
+                  Schema.Field.of(jsonBatchKey, bufferRecordArraySchema));
+          StructuredRecord wrappedMessageRecord = StructuredRecord.builder(wrappedMessegeSchema)
+                  .set(jsonBatchKey, buffer).build();
+          message.append(StructuredRecordStringConverter.toJsonString(wrappedMessageRecord));
```
albertshau: the message StringBuilder should be a variable local to this method. Otherwise calling this method will have side effects. For example, calling it twice in a row without calling clear() would double the message.

src/main/java/io/cdap/plugin/http/sink/batch/MessageBuffer.java
psainics: Updated

src/main/java/io/cdap/plugin/http/sink/batch/MessageBuffer.java
psainics: moved to local scope

src/main/java/io/cdap/plugin/http/sink/batch/MessageBuffer.java
psainics: now that writer class checks isEmpty before calling this method, it is no longer required.

src/main/java/io/cdap/plugin/http/sink/batch/MessageBuffer.java
psainics: Updated

src/main/java/io/cdap/plugin/http/sink/batch/HTTPRecordWriter.java
psainics: removed

src/main/java/io/cdap/plugin/http/sink/batch/MessageBuffer.java
psainics: Updated

src/main/java/io/cdap/plugin/http/sink/batch/HTTPSinkConfig.java
psainics: Updated

src/main/java/io/cdap/plugin/http/sink/batch/HTTPSinkConfig.java
psainics: Updated

docs/HTTP-batchsink.md
psainics: Updated

src/main/java/io/cdap/plugin/http/sink/batch/MessageBuffer.java
```diff
@@ -0,0 +1,203 @@
...
+    this.jsonBatchKey = jsonBatchKey;
+    this.delimiterForMessages = delimiterForMessages;
+    this.charset = charset;
+    this.shouldWriteJsonAsArray = shouldWriteJsonAsArray;
+    this.customMessageBody = customMessageBody;
+    this.buffer = new ArrayList<>();
+    switch (messageFormat.toLowerCase()) {
+      case "json":
+        messageFormatter = this::formatAsJsonNoException;
+        contentType = contentTypeMap.get("json");
```
albertshau: nit: don't need the map anymore, can just assign the proper string in each switch case.

src/main/java/io/cdap/plugin/http/sink/batch/MessageBuffer.java
```diff
@@ -0,0 +1,203 @@
...
+      throw new IllegalStateException("Error formatting JSON message. Reason: " + e.getMessage(), e);
+    }
+  }
+
+  private String formatAsJson(List<StructuredRecord> buffer) throws IOException {
+    boolean useJsonBatchKey = !Strings.isNullOrEmpty(jsonBatchKey);
+    if (!shouldWriteJsonAsArray) {
+      return String.join(delimiterForMessages, getBufferAsJsonList());
+    }
+    if (useJsonBatchKey) {
```
albertshau: nit: more readable to put the short conditions at the start and then remove nesting for the larger code block:

```
if (!shouldWriteJsonAsArray) {
  return ...
}
if (!useJsonBatchKey) {
  return ...
} 
Schema bufferRecordSchema = ..
...
```

src/main/java/io/cdap/plugin/http/sink/batch/MessageBuffer.java
```diff
@@ -0,0 +1,203 @@
...
+      // A new StructuredRecord is created with the jsonBatchKey as the
+      // field name and the array of records as the value
+      Schema bufferRecordSchema = buffer.get(0).getSchema();
+      Schema bufferRecordArraySchema = Schema.arrayOf(bufferRecordSchema);
+      Schema wrappedMessegeSchema = Schema.recordOf("wrapper",
```
albertshau: nit: typo Messege -> Message.

Also, should create this schema once in the constructor instead of here for each method call.

src/main/java/io/cdap/plugin/http/sink/batch/MessageBuffer.java
```diff
@@ -0,0 +1,203 @@
...
+  private final Function<List<StructuredRecord>, String> messageFormatter;
+  private final String contentType;
+  private static final Map<String, String> contentTypeMap = new HashMap<String, String>() {{
+    put("json", "application/json");
+    put("form", "application/x-www-form-urlencoded");
+    put("custom", "text/plain");
+  }};
+
+  public MessageBuffer(
+          String messageFormat, String jsonBatchKey, Boolean shouldWriteJsonAsArray,
```
albertshau: this should be a boolean instead of Boolean to avoid potential null pointer exceptions

src/main/java/io/cdap/plugin/http/sink/batch/MessageBuffer.java
```diff
@@ -0,0 +1,203 @@
...
+      // field name and the array of records as the value
+      Schema bufferRecordSchema = buffer.get(0).getSchema();
+      Schema bufferRecordArraySchema = Schema.arrayOf(bufferRecordSchema);
+      Schema wrappedMessegeSchema = Schema.recordOf("wrapper",
+              Schema.Field.of(jsonBatchKey, bufferRecordArraySchema));
+      StructuredRecord wrappedMessageRecord = StructuredRecord.builder(wrappedMessegeSchema)
+              .set(jsonBatchKey, buffer).build();
+      return StructuredRecordStringConverter.toJsonString(wrappedMessageRecord);
+    } else {
+      return "[" + String.join(",", getBufferAsJsonList()) + "]";
```
albertshau: still have the comment about doing this in one loop instead of two. For example, can change getBufferAsJsonList() to take a StringBuilder as an argument and append to it as needed.

psainics: This will do the work in one loop
```java
private String getBufferAsJsonList() throws IOException {
    StringBuilder sb = new StringBuilder();
    int size = buffer.size();
    String delimiter = shouldWriteJsonAsArray ? "," : delimiterForMessages;
    if (shouldWriteJsonAsArray) {
      sb.append("[");
    }
    for (StructuredRecord record : buffer) {
      sb.append(StructuredRecordStringConverter.toJsonString(record));
      if (--size > 0) {
        sb.append(delimiter);
      }
    }
    if (shouldWriteJsonAsArray) {
      sb.append("]");
    }
    return sb.toString();
  }
  ```

File: src/main/java/io/cdap/plugin/http/sink/batch/MessageBuffer.java (hunk 32 above)
Comment (psainics) [reply src/main/java/io/cdap/plugin/http/sink/batch/MessageBuffer.java]: updated

File: src/main/java/io/cdap/plugin/http/sink/batch/MessageBuffer.java (hunk 31 above)
Comment (psainics) [reply src/main/java/io/cdap/plugin/http/sink/batch/MessageBuffer.java]: Fixed

File: src/main/java/io/cdap/plugin/http/sink/batch/MessageBuffer.java (hunk 30 above)
Comment (psainics) [reply src/main/java/io/cdap/plugin/http/sink/batch/MessageBuffer.java]: Updated

File: src/main/java/io/cdap/plugin/http/sink/batch/MessageBuffer.java (hunk 29 above)
Comment (psainics) [reply src/main/java/io/cdap/plugin/http/sink/batch/MessageBuffer.java]: Removed map

File: src/main/java/io/cdap/plugin/http/sink/batch/MessageBuffer.java (hunk 31 above)
Comment (psainics) [reply src/main/java/io/cdap/plugin/http/sink/batch/MessageBuffer.java]: We need at least 1 element in the buffer to create this schema.

Comment (psainics) [reply src/main/java/io/cdap/plugin/http/sink/batch/MessageBuffer.java]: We need at least 1 element in the buffer to create this schema.

File: src/main/java/io/cdap/plugin/http/sink/batch/MessageBuffer.java [hunk 34]
```diff
@@ -0,0 +1,205 @@
...
+      case "custom":
+        messageFormatter = this::formatAsCustom;
+        contentType = "text/plain";
+        break;
+      default:
+        throw new IllegalArgumentException("Invalid message format: " + messageFormat);
+    }
+  }
+
+  public void add(StructuredRecord record) {
```
DJSagarAhire, line 102: Javadocs are missing for all the public methods in the class.

src/main/java/io/cdap/plugin/http/sink/batch/MessageBuffer.java
```diff
@@ -0,0 +1,205 @@
...
+
+  public String getContentType() {
+    return contentType;
+  }
+
+  public String getMessage() throws IOException {
+    return messageFormatter.apply(buffer);
+  }
+
+  private String formatAsJsonNoException(List<StructuredRecord> buffer) {
```
DJSagarAhire: The method name confuses the purpose of the method. Ideally the method name that is to be used throughout the class should hide the implementation details, so I'd suggest changing this method to `formatAsJson()` and renaming the other one to something else.

src/main/java/io/cdap/plugin/http/sink/batch/MessageBuffer.java
```diff
@@ -0,0 +1,205 @@
...
+          String messageFormat, String jsonBatchKey, boolean shouldWriteJsonAsArray,
+          String delimiterForMessages, String charset, String customMessageBody
+  ) {
+    this.jsonBatchKey = jsonBatchKey;
+    this.delimiterForMessages = delimiterForMessages;
+    this.charset = charset;
+    this.shouldWriteJsonAsArray = shouldWriteJsonAsArray;
+    this.customMessageBody = customMessageBody;
+    this.buffer = new ArrayList<>();
+    switch (messageFormat.toLowerCase()) {
```
DJSagarAhire: Might be valuable to make `messageFormat` an enum instead.

src/main/java/io/cdap/plugin/http/sink/batch/MessageBuffer.java
psainics: Added Javadocs for public methods

src/main/java/io/cdap/plugin/http/sink/batch/MessageBuffer.java
psainics: Renamed to `formatAsJson` and the other to `formatAsJsonInternal`

src/main/java/io/cdap/plugin/http/sink/batch/MessageBuffer.java
psainics: Added MessageFormatType

psainics:
@albertshau can you review the changes.

src/main/java/io/cdap/plugin/http/common/http/MessageFormatType.java
```diff
@@ -0,0 +1,43 @@
...
+import io.cdap.plugin.http.common.EnumWithValue;
+
+/**
+ * An enum which represent a type of message format of an HTTP request body.
+ */
+public enum MessageFormatType implements EnumWithValue {
+
+  JSON("JSON"),
+  FORM("FORM"),
+  CUSTOM("CUSTOM");
```
albertshau: there isn't a need for these values, can just use a plain enum. `Enum.name()` is the same thing as these values.

src/main/java/io/cdap/plugin/http/sink/batch/MessageBuffer.java
```diff
@@ -0,0 +1,232 @@
...
+  private String formatAsJsonInternal(List<StructuredRecord> buffer) throws IOException {
+    boolean useJsonBatchKey = !Strings.isNullOrEmpty(jsonBatchKey);
+    if (!shouldWriteJsonAsArray || !useJsonBatchKey) {
+      return getBufferAsJsonList();
+    }
+    // A new StructuredRecord is created with the jsonBatchKey as the
+    // field name and the array of records as the value
+    Schema bufferRecordSchema = buffer.get(0).getSchema();
+    Schema bufferRecordArraySchema = Schema.arrayOf(bufferRecordSchema);
+    Schema wrappedMessageSchema = Schema.recordOf("wrapper",
```
albertshau: still have the comment that this schema should be created once in the constructor instead of for every method call

src/main/java/io/cdap/plugin/http/sink/batch/MessageBuffer.java
```diff
@@ -0,0 +1,232 @@
...
+  private String getBufferAsJsonList() throws IOException {
+    StringBuilder sb = new StringBuilder();
+    int size = buffer.size();
+    String delimiter = shouldWriteJsonAsArray ? "," : delimiterForMessages;
+    if (shouldWriteJsonAsArray) {
+      sb.append("[");
+    }
+    for (StructuredRecord record : buffer) {
+      sb.append(StructuredRecordStringConverter.toJsonString(record));
+      if (--size > 0) {
```
albertshau: nit: since we're using a StringBuilder, it's more efficient to remove the last character at the end of the loop (assuming at least 1 element in the buffer) rather than doing this check every time through the loop.

albertshau:
just a few minor comments, please fix and squash commits, then we can merge

src/main/java/io/cdap/plugin/http/sink/batch/MessageBuffer.java
psainics:
```java
if (!buffer.isEmpty()) {
  sb.setLength(sb.length() - delimiter.length());
}
 ```

File: src/main/java/io/cdap/plugin/http/common/http/MessageFormatType.java (hunk 37 above)
Comment (psainics) [reply src/main/java/io/cdap/plugin/http/common/http/MessageFormatType.java]:
```java
public enum MessageFormatType implements EnumWithValue {

  JSON,
  FORM,
  CUSTOM;

  @Override
  public String getValue() {
    return name();
  }

  @Override
  public String toString() {
    return this.getValue();
  }
}
```

src/main/java/io/cdap/plugin/http/sink/batch/MessageBuffer.java
psainics: Moved to constructor

DJSagarAhire (approved):
LGTM. Do address @albertshau's comments and squash the commits before merging.

PR #152, opened 2024-01-22
[PLUGIN-1737] Added Schema Detection for CSV/TSV format

psainics:
##  Added Schema Detection for CSV/TSV format 

Jira : [Plugin-1737](https://cdap.atlassian.net/browse/PLUGIN-1737)

## Description
User will be able to get Schema for delimited files (CSV/TSV)

## UI Fields changes
- Add Schema Button appears when format is csv/tsv
- Sample Size filed appears when format is csv/tsv
- `CSV Skip First Row` is renamed to `CSV/TSV Skip First Row`
- Added new UI field `Enable Quoted Values`

image

## Docs
 - Docs for new fileds added.
 - CSV Skip First Row renamed.
---

**Get Schema:** Auto-detects schema from file. Supported formats are: csv, tsv.
**Sample Size:** The maximum number of rows in a file that will get investigated for automatic data type detection. 
Default is 100.
**CSV/TSV Skip First Row:** Whether to skip the first row of the HTTP response. This is usually set if the first row is a header row.

---
## Unit Tests

- Unit Test for `DelimitedSchemaDetector`

image

- Unit Test for `DelimitedPage`

image

- Unit Test for `RawStringPerLine`

image

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -84,6 +84,7 @@
     <hadoop.version>2.3.0</hadoop.version>
     <httpcomponents.version>4.5.9</httpcomponents.version>
     <hydrator.version>2.10.0-SNAPSHOT</hydrator.version>
+    <cdap.plugin.version>2.12.0</cdap.plugin.version>
     <jackson.version>2.9.9</jackson.version>
     <junit.version>4.11</junit.version>
     <jython.version>2.7.1</jython.version>
@@ -125,6 +126,11 @@
       <groupId>io.cdap.plugin</groupId>
       <artifactId>hydrator-common</artifactId>
       <version>${hydrator.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>io.cdap.plugin</groupId>
+      <artifactId>format-delimited</artifactId>
+      <version>${cdap.plugin.version}</version>
     </dependency>
     <dependency>
       <groupId>org.apache.hadoop</groupId>
--- a/src/main/java/io/cdap/plugin/http/common/pagination/page/DelimitedPage.java
+++ b/src/main/java/io/cdap/plugin/http/common/pagination/page/DelimitedPage.java
@@ -16,12 +16,17 @@
 
 package io.cdap.plugin.http.common.pagination.page;
 
+import com.google.common.base.Splitter;
 import io.cdap.cdap.api.data.format.StructuredRecord;
-import io.cdap.cdap.format.StructuredRecordStringConverter;
+import io.cdap.cdap.api.data.schema.Schema;
+import io.cdap.plugin.format.delimited.common.DelimitedStructuredRecordStringConverter;
+import io.cdap.plugin.format.delimited.input.SplitQuotesIterator;
 import io.cdap.plugin.http.common.http.HttpResponse;
 import io.cdap.plugin.http.source.common.BaseHttpSourceConfig;
 
+
 import java.io.IOException;
+import java.util.Iterator;
 import java.util.NoSuchElementException;
 
 /**
@@ -39,7 +44,14 @@ public class DelimitedPage extends RecordPerLinePage {
 
   @Override
   protected StructuredRecord getStructedRecordByString(String line) {
-    return StructuredRecordStringConverter.fromDelimitedString(line, delimiter, schema);
+    StructuredRecord.Builder builder = StructuredRecord.builder(schema);
+    Iterator<String> splitsIterator = getSplitsIterator(config.getEnableQuotesValues(), line, delimiter);
+    Iterator<Schema.Field> fields = schema.getFields().iterator();
+    while (splitsIterator.hasNext()) {
+      Schema.Field field = fields.next();
+      DelimitedStructuredRecordStringConverter.parseAndSetFieldValue(builder, field, splitsIterator.next());
+    }
+    return builder.build();
   }
 
   @Override
@@ -52,5 +64,13 @@ public class DelimitedPage extends RecordPerLinePage {
       isLineRead = false;
     }
     return super.next();
+  }
+
+  private Iterator<String> getSplitsIterator(boolean enableQuotesValue, String delimitedString, String delimiter) {
+    if (enableQuotesValue) {
+      return new SplitQuotesIterator(delimitedString, delimiter, null, false);
+    } else {
+      return Splitter.on(delimiter).split(delimitedString).iterator();
+    }
   }
 }
--- a/src/main/java/io/cdap/plugin/http/source/batch/HttpInputFormatProvider.java
+++ b/src/main/java/io/cdap/plugin/http/source/batch/HttpInputFormatProvider.java
@@ -20,8 +20,17 @@ import com.google.common.collect.ImmutableMap;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import io.cdap.cdap.api.data.batch.InputFormatProvider;
+import io.cdap.cdap.api.data.schema.Schema;
+import io.cdap.cdap.etl.api.FailureCollector;
+import io.cdap.plugin.http.common.http.HttpClient;
+import io.cdap.plugin.http.common.http.HttpResponse;
+import io.cdap.plugin.http.common.pagination.page.PageFormat;
+import io.cdap.plugin.http.source.common.DelimitedSchemaDetector;
+import io.cdap.plugin.http.source.common.RawStringPerLine;
 
+import java.io.IOException;
 import java.util.Map;
+import javax.annotation.Nullable;
 
 /**
  * InputFormatProvider used by cdap to provide configurations to mapreduce job
@@ -31,11 +40,13 @@ public class HttpInputFormatProvider implements InputFormatProvider {
   private static final Gson gson = new GsonBuilder().create();
 
   private final Map<String, String> conf;
+  private final HttpBatchSourceConfig config;
 
   HttpInputFormatProvider(HttpBatchSourceConfig config) {
     this.conf = new ImmutableMap.Builder<String, String>()
       .put(PROPERTY_CONFIG_JSON, gson.toJson(config))
       .build();
+    this.config = config;
   }
 
   @Override
@@ -46,5 +57,26 @@ public class HttpInputFormatProvider implements InputFormatProvider {
   @Override
   public Map<String, String> getInputFormatConfiguration() {
     return conf;
+  }
+
+  @Nullable
+  public Schema getSchema(FailureCollector failureCollector) {
+    PageFormat format = config.getFormat();
+    switch (format) {
+      case CSV:
+      case TSV:
+        String delimiter = format == PageFormat.CSV ? "," : "\t";
+        try (HttpClient client = new HttpClient(config)) {
+          RawStringPerLine rawStringPerLine = new RawStringPerLine(
+            new HttpResponse(client.executeHTTP(config.getUrl())));
+          return DelimitedSchemaDetector.detectSchema(config, delimiter, rawStringPerLine, failureCollector);
+        } catch (IOException e) {
+          failureCollector.addFailure(String.format("Error while reading the file to infer the schema. Error: %s",
+                                                   e.getMessage()), null);
+        }
+        return null;
+      default:
+        return null;
+    }
   }
 }
--- a/src/main/java/io/cdap/plugin/http/source/common/BaseHttpSourceConfig.java
+++ b/src/main/java/io/cdap/plugin/http/source/common/BaseHttpSourceConfig.java
@@ -24,21 +24,17 @@ import io.cdap.cdap.etl.api.FailureCollector;
 import io.cdap.cdap.etl.api.validation.InvalidConfigPropertyException;
 import io.cdap.cdap.etl.api.validation.InvalidStageException;
 import io.cdap.plugin.common.ReferenceNames;
-import io.cdap.plugin.common.ReferencePluginConfig;
 import io.cdap.plugin.http.common.BaseHttpConfig;
 import io.cdap.plugin.http.common.EnumWithValue;
 import io.cdap.plugin.http.common.RetryPolicy;
 import io.cdap.plugin.http.common.error.ErrorHandling;
 import io.cdap.plugin.http.common.error.HttpErrorHandlerEntity;
 import io.cdap.plugin.http.common.error.RetryableErrorHandling;
-import io.cdap.plugin.http.common.http.AuthType;
 import io.cdap.plugin.http.common.http.KeyStoreType;
-import io.cdap.plugin.http.common.http.OAuthUtil;
 import io.cdap.plugin.http.common.pagination.PaginationIteratorFactory;
 import io.cdap.plugin.http.common.pagination.PaginationType;
 import io.cdap.plugin.http.common.pagination.page.PageFormat;
 
-import java.io.File;
 import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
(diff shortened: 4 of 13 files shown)


src/main/java/io/cdap/plugin/http/common/pagination/page/DelimitedStructuredRecordStringConverter.java
```diff
@@ -0,0 +1,49 @@
...
+   * @param builder Structured Record Builder instance
+   * @param field   field to set
+   * @param part    String portion of the delimited input
+   */
+  public static void parseAndSetFieldValue(StructuredRecord.Builder builder,
+                                           Schema.Field field,
+                                           String part) {
+    if (part.isEmpty()) {
+      builder.set(field.getName(), null);
+    } else {
```
albertshau: nit: in these situations, it's more readable to have the edge conditions checks at the start of the method with returns within each condition. Then have the main logic afterwards :

```
if (...) {
  // something
  return;
}
// other logic
```

src/main/java/io/cdap/plugin/http/source/batch/HttpInputFormatProvider.java
```diff
@@ -47,4 +54,19 @@ public String getInputFormatClassName() {
   public Map<String, String> getInputFormatConfiguration() {
     return conf;
   }
```
albertshau, line 60: newline between methods

src/main/java/io/cdap/plugin/http/source/batch/HttpInputFormatProvider.java
```diff
@@ -47,4 +54,19 @@ public String getInputFormatClassName() {
...
+  public Schema getSchema(FailureCollector failureCollector) {
+    PageFormat format = config.getFormat();
+    switch (format) {
+      case CSV:
+      case TSV:
+        String delimiter = format == PageFormat.CSV ? "," : "\t";
+        return DelimitedSchemaDetector.detectSchema(config, delimiter, failureCollector);
+      default:
+        failureCollector.addFailure(String.format(
+                "%s format cannot be used without specifying a schema. Schema must be specified.", format), null)
```
albertshau: this is an odd error message, since the schema isn't checked anywhere here. Is this method supposed to be returning the configured schema if it is not null?

src/main/java/io/cdap/plugin/http/source/common/BaseHttpSourceConfig.java
```diff
@@ -676,18 +695,84 @@ PAGINATION_INDEX_PLACEHOLDER, getPaginationType()),
     }
   }
 
+  public boolean canDetectSchema() {
+    // Schema and format are not macros, so we can detect schema
+    if (containsMacro(PROPERTY_SCHEMA) || containsMacro(PROPERTY_FORMAT)) {
+      return false;
+    }
+    // If url is macro, we cannot detect schema
+    if (containsMacro(PROPERTY_URL)) {
```
albertshau: this should be combined with the previous if condition

src/main/java/io/cdap/plugin/http/source/common/BaseHttpSourceConfig.java
```diff
@@ -676,18 +695,84 @@ PAGINATION_INDEX_PLACEHOLDER, getPaginationType()),
...
+        }
+        break;
+      case SERVICE_ACCOUNT:
+        if (containsMacro(PROPERTY_NAME_SERVICE_ACCOUNT_FILE_PATH)) {
+          return false;
+        }
+        break;
+    }
+    // Check if url is https and verify https is macro
+    if (getUrl().startsWith("https")) {
```
albertshau: the logic in this method is quite complex, but it seems to be logically the OR of all of these properties. Is there a reason we're breaking it up into a bunch of different if conditions?

src/main/java/io/cdap/plugin/http/source/common/DataTypeDetectorUtils.java
```diff
@@ -0,0 +1,206 @@
...
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Utils facilitating automated schema detection.
+ */
+public class DataTypeDetectorUtils {
```
albertshau: are all the DataType classes copied from https://github.com/cdapio/hydrator-plugins/blob/develop/format-delimited/src/main/java/io/cdap/plugin/format/delimited/common/DataType.java? Or are there logical differences.

If it's the same, we should be adding that library as a dependency rather than duplicating code. Otherwise the logic will slowly diverge over time.

albertshau (commented):
left a few comments, but there seems to be a lot of copied code so it's hard to tell what needs to be reviewed in more detail. Copy-pasting code is almost never the right approach, please switch to use the existing code instead of duplicating it.

src/main/java/io/cdap/plugin/http/common/pagination/page/DelimitedStructuredRecordStringConverter.java
psainics: This class is removed using it from the library now.

src/main/java/io/cdap/plugin/http/source/batch/HttpInputFormatProvider.java
psainics: Fixed!

src/main/java/io/cdap/plugin/http/source/batch/HttpInputFormatProvider.java
psainics: Yup, this is not required here, this has been remove.

src/main/java/io/cdap/plugin/http/source/common/BaseHttpSourceConfig.java
psainics: Fixed.

src/main/java/io/cdap/plugin/http/source/common/BaseHttpSourceConfig.java
psainics: I was trying to handle cases where say a user has accidentally put a SSL/TLS config field as macro but if the url is not `https` we can still get the schema.
I broke up fields so they can be grouped as they appear in UI sections. 

I believe this is not required, i have combined all the fields in logical OR now.

src/main/java/io/cdap/plugin/http/source/common/DataTypeDetectorUtils.java
psainics: Added dependency.

psainics:
@albertshau The duplicated code is remove. I was not aware the `format-X` are published deps that can use outside the hydrator repo.

Can you help me with which `cdap.plugin.version` is appropriate for use here currently i have added the latest `=2.13.0-SNAPSHOT`.

pom.xml
```diff
@@ -84,6 +84,7 @@
     <hadoop.version>2.3.0</hadoop.version>
     <httpcomponents.version>4.5.9</httpcomponents.version>
     <hydrator.version>2.10.0-SNAPSHOT</hydrator.version>
+    <cdap.plugin.version>2.13.0-SNAPSHOT</cdap.plugin.version>
```
albertshau: should use the latest non-snapshot version (would guess it's a 2.12.x version)

src/main/java/io/cdap/plugin/http/common/http/HttpClient.java
```diff
@@ -53,8 +54,18 @@ public class HttpClient implements Closeable {
...
   public HttpClient(BaseHttpSourceConfig config) {
+    this(config, Collections.emptyMap());
+  }
+
+  public HttpClient(BaseHttpSourceConfig config, Map<String, String> extraHeaders) {
     this.config = config;
-    this.headers = config.getHeadersMap();
+    // merge headers from config and extra headers. Extra headers will override headers from config
+    if (config.getHeadersMap() != null) {
+      this.headers = config.getHeadersMap();
```
albertshau: it's better to make a copy of the map `new HashMap<>(config.getHeadersMap())`. Otherwise this would actually modify the map from the config, which could lead to weird behavior.

On a separate note, normally it's good to return an unmodifiable map from getter methods to prevent this type of issue.

src/main/java/io/cdap/plugin/http/common/http/HttpClient.java
```diff
@@ -53,8 +54,18 @@ public class HttpClient implements Closeable {
...
+      this.headers.putAll(extraHeaders);
+    } else {
+      this.headers = extraHeaders;
```
albertshau: similar comment here, make a copy of the map

src/main/java/io/cdap/plugin/http/common/pagination/page/DelimitedPage.java
```diff
@@ -53,4 +65,11 @@ public PageEntry next() {
     }
     return super.next();
   }
```
albertshau, line 67: nit: add newline between methods

src/main/java/io/cdap/plugin/http/source/common/DelimitedSchemaDetector.java
```diff
@@ -0,0 +1,64 @@
...
+import io.cdap.plugin.format.delimited.common.DataTypeDetectorUtils;
+import io.cdap.plugin.http.source.batch.HttpBatchSourceConfig;
+
+import java.util.HashMap;
+import java.util.List;
+
+/**
+ * Class that detects the schema of the delimited file.
+ */
+public class DelimitedSchemaDetector {
```
albertshau, line 31: can you add a unit test for this class?

src/main/java/io/cdap/plugin/http/source/common/DelimitedSchemaDetector.java
```diff
@@ -0,0 +1,64 @@
...
+  public static Schema detectSchema(HttpBatchSourceConfig config, String delimiter, FailureCollector failureCollector) {
+    HttpPartialContentRequest partialContentRequest = new HttpPartialContentRequest(config);
+    HttpPartialContentReader partialContentReader = new HttpPartialContentReader(partialContentRequest);
```
albertshau: Is there no existing class that we can re-use for reading the http content? It seems odd to have a entirely new class with different logic that may potentially read things differently than the reader for the actual data.

pom.xml
psainics: Using [2.12.0](https://mvnrepository.com/artifact/io.cdap.plugin/format-common/2.12.0)

src/main/java/io/cdap/plugin/http/common/http/HttpClient.java
psainics: Okay, fixed.
Also i have changed the getter to return an unmodifiableMap.

src/main/java/io/cdap/plugin/http/common/http/HttpClient.java
psainics: Fixed

src/main/java/io/cdap/plugin/http/common/pagination/page/DelimitedPage.java
psainics: Added !

src/main/java/io/cdap/plugin/http/source/common/DelimitedSchemaDetector.java
psainics: Added test cases.
image

src/main/java/io/cdap/plugin/http/source/common/DelimitedSchemaDetector.java
psainics: The problem with the current HTTP reader class is that it performs a basic `GET` request and reads the entire file.

This approach becomes highly inefficient, especially for larger files (CSV/TSV), as it unnecessarily consumes bandwidth. To address this issue, I have developed a new class that allows for partial file reading. This class abstracts the process of reading files partially and provides a getLine() method, which continuously returns the next line of the file.

albertshau: I don't think this partial one works properly though. For example, it doesn't look like it has implemented any of the logic around oauth. I also don't understand how it would be more efficient, as it looks like it is creating a new HttpClient and making a new request for each chunk. Even if it were doing everything correctly, it would be by duplicating logic, which would easily get out of sync in the future.

I took a look at the existing logic, it doesn't look like it is doing anything terribly inefficient. RecordPerLinePage is opening an InputStream from the HTTP response and reading it line by line. We should be able to read N lines and close the reader without issue.

psainics: Why do you suggest this does not work with OAuth 🤔 
I am still using the wrapper to HTTP provided, that abstracts all the proxy and auth logic.

> We should be able to read N lines and close the reader without issue.

Yes this is possible, the class `RecordPerLinePage` does not directly have a methord to return raw lines (which we require) but a new class can be created that extends `RecordPerLinePage`.

One thing also discussed was user accidentally using a wrong or corrupted file with no `\n`.
The buffer-reader will try to read complete file but here we have implement a 10 consumptive chunk loading limit to prevent such issues.

Other that that there is no specific reason to use this class.
What's your suggestion ?

albertshau: I see, thought HttpClient was a library, but it's part of this project, so ignore that comment about OAuth.

I still think its better to re-use the existing classes as much as possible to read line by line. This ensures that the sampling logic is as similar as possible to the actual reading logic. For example, what if the server behind the url doesn't honor the content range headers? Or what if the number of samples is set to a number higher than the results returned in a single page? In that case, it seems like we would want to make requests for additional pages.

I don't think we need to worry too much about the no newline edge case. Setting a 10 chunk limit while sampling but not while reading again makes the behavior inconsistent. If it is really a concern it should be handled in a separate PR, perhaps as an advanced configuration property that applies both to sampling and reading.

psainics: I have created a new class `RawStringPerLine`, that is a iterator of String, and takes HTTP response as an input.`

src/main/java/io/cdap/plugin/http/source/common/RawStringPerLine.java
```diff
@@ -0,0 +1,76 @@
...
+    protected final HttpResponse httpResponse;
+    private BufferedReader bufferedReader;
+    protected boolean isLineRead;
+    private String lastLine;
+
+    public RawStringPerLine(HttpResponse httpResponse) {
+        this.httpResponse = httpResponse;
+    }
+
+    BufferedReader getBufferedReader() throws IOException {
```
albertshau: this can be private

src/main/java/io/cdap/plugin/http/source/common/RawStringPerLine.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+
+/**
+ * Class that reads the raw string from the HTTP response and returns it line by line.
+ */
+public class RawStringPerLine implements Closeable, Iterator<String> {
+    protected final HttpResponse httpResponse;
+    private BufferedReader bufferedReader;
+    protected boolean isLineRead;
```
albertshau: this can be private, unless there's a subclass that I'm missing

src/main/java/io/cdap/plugin/http/source/common/RawStringPerLine.java
```diff
@@ -0,0 +1,76 @@
...
+import java.io.Closeable;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+
+/**
+ * Class that reads the raw string from the HTTP response and returns it line by line.
+ */
+public class RawStringPerLine implements Closeable, Iterator<String> {
```
albertshau, line 31: this class still has some of the issues mentioned before, like it won't honor the number of samples if the first page has fewer rows. 

Let's save that for a follow up though.

widgets/HTTP-batchsource.json
```diff
@@ -68,6 +68,19 @@
             "default": "json"
           }
         },
+        {
+          "widget-type": "number",
+          "label": "Sample Size",
+          "name": "sampleSize",
+          "widget-attributes": {
+            "default": "1000",
```
albertshau: maybe better to default it to a lower number like 100. 1000 is often much more than the number of results in a single page.

albertshau (approved):
lgtm

albertshau:
minor comments, please fix and squash commits, then we can merge

widgets/HTTP-batchsource.json
psainics: Changed all occurrence to default to 100;

src/main/java/io/cdap/plugin/http/source/common/RawStringPerLine.java
psainics: Done !

src/main/java/io/cdap/plugin/http/source/common/RawStringPerLine.java
psainics: Done !

PR #158, opened 2024-03-06
[PLUGIN-1756] Fixed URL Validation

psainics:
## Fixed URL Validation

Jira : [PLUGIN-1756](https://cdap.atlassian.net/browse/PLUGIN-1756)

### Description

- When an invalid url is given, the validation logic only checks for a valid URI which may not raise errors for a lot of strings that are not valid URLs.
Replacing the validation to check for a URL will fix the proper error.

- Some exceptions are thrown by the http library without any error message string.
This is usually present on the cause of the error.

Adding this code snippet will fix null issue for other exceptions thrown by the library.

```java
String errorMessage = e.getMessage();
if (errorMessage == null && e.getCause() != null) {
errorMessage = e.getCause().getMessage();
}
```

### UI Field

- No Changes made to widget json.

### Docs

- No Changes made to docs.

### Code change

- Modified `HttpInputFormatProvider.java`
- Modified `BaseHttpSourceConfig.java`

### Unit Tests

- No Changes made to unit tests.

### DIFF
--- a/src/main/java/io/cdap/plugin/http/source/batch/HttpInputFormatProvider.java
+++ b/src/main/java/io/cdap/plugin/http/source/batch/HttpInputFormatProvider.java
@@ -71,8 +71,12 @@ public class HttpInputFormatProvider implements InputFormatProvider {
             new HttpResponse(client.executeHTTP(config.getUrl())));
           return DelimitedSchemaDetector.detectSchema(config, delimiter, rawStringPerLine, failureCollector);
         } catch (IOException e) {
+          String errorMessage = e.getMessage();
+          if (Strings.isNullOrEmpty(errorMessage) && e.getCause() != null) {
+            errorMessage = e.getCause().getMessage();
+          }
           failureCollector.addFailure(String.format("Error while reading the file to infer the schema. Error: %s",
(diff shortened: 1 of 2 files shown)


src/main/java/io/cdap/plugin/http/source/batch/HttpInputFormatProvider.java
```diff
@@ -71,8 +71,12 @@ public Schema getSchema(FailureCollector failureCollector) {
             new HttpResponse(client.executeHTTP(config.getUrl())));
           return DelimitedSchemaDetector.detectSchema(config, delimiter, rawStringPerLine, failureCollector);
         } catch (IOException e) {
+          String errorMessage = e.getMessage();
+          if (errorMessage == null && e.getCause() != null) {
```
MrRahulSharma:
```suggestion
          if (Strings.isNullOrEmpty(errorMessage) && e.getCause() != null) {
```

psainics: Changes Accepted !

MrRahulSharma (approved):

PR #159, opened 2024-03-07
[PLUGIN-1757] Remove Page iteration from transform

psainics:
##  Remove Page iteration from transform

Jira : [PLUGIN-1757](https://cdap.atlassian.net/browse/PLUGIN-1757)

### Description

Plugin had very high memory usage compared to the input file size.

This PR improves memory usage by removing Page iteration from transform.

### OLD
image

### NEW
image

### UI Field

- No Changes made to widget json.

### Docs

- No Changes made to docs.

### Code change

- Modified `HttpBatchSource.java`
- Modified `HttpInputFormat.java`
- Modified `HttpRecordReader.java`

### Unit Tests

- No Changes made to unit tests.

### Results

- This PR ( Before removing page iteration and current implementation )
image

- This PR with closing the basePage
image

### DIFF
--- a/src/main/java/io/cdap/plugin/http/source/batch/HttpRecordReader.java
+++ b/src/main/java/io/cdap/plugin/http/source/batch/HttpRecordReader.java
@@ -20,6 +20,7 @@ import com.google.gson.GsonBuilder;
 import io.cdap.plugin.http.common.pagination.BaseHttpPaginationIterator;
 import io.cdap.plugin.http.common.pagination.PaginationIteratorFactory;
 import io.cdap.plugin.http.common.pagination.page.BasePage;
+import io.cdap.plugin.http.common.pagination.page.PageEntry;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.io.NullWritable;
 import org.apache.hadoop.mapreduce.InputSplit;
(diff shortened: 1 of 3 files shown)


MrRahulSharma:
Can you add the testing details performed for this change ?

psainics:
> Can you add the testing details performed for this change ?

For performance i have tested this with a 2GB csv file.
We saw memory usage reduced from 7GB to 1.5GB.

I have done pagination test to make sure it's same.

Testing for other formats is in being done !

MrRahulSharma (approved):

src/main/java/io/cdap/plugin/http/source/batch/HttpRecordReader.java
```diff
@@ -57,10 +59,22 @@ public void initialize(InputSplit inputSplit, TaskAttemptContext taskAttemptCont
...
-      return false;
+    // If there is no current page or no next line in the current page
+    if (currentBasePage == null || !currentBasePage.hasNext()) {
+      if (!httpPaginationIterator.hasNext()) {
+        // If there is no next page, return false
+        // All pages are read
+        return false;
+      }
+      // Get the next page
+      currentBasePage = httpPaginationIterator.next();
```
arjan-bal, line 70: We are not closing the currentBasePage here. This [can hold a BufferedInputReader](https://github.com/data-integrations/http/blob/557bea0b1dd31351b171572e13e428dd9bbd8cc5/src/main/java/io/cdap/plugin/http/common/pagination/page/RecordPerLinePage.java#L85-L89). If we don't close it, its memory may still be freed when it is finalized.

Can you try closing the page to see if it improves the memory utilization further?

vikasrathee-cs: We have looked into this and closing the basepage is not making any impact on the memory usage, but its better to close it and we will do the same in next release. Adding snapshot in the description for memory usage after closing the reader.

MrRahulSharma: Good catch. We should make sure that we do close it.

PR #165, opened 2024-04-10
[PLUGIN-1775]Changes done for schema bug while using enable quotes

vikasrathee-cs:
[PLUGIN-1775] Changes done for schema detection bug while using enable quotes with csv format.


src/main/java/io/cdap/plugin/http/source/common/DelimitedSchemaDetector.java
```diff
@@ -61,4 +64,23 @@ public static Schema detectSchema(HttpBatchSourceConfig config, String delimiter
...
   }
+
+  /**
+   * @param rawLine            line to parse and find out the exact number of columns in a row.
+   * @param enableQuotedValues flag whether file can contain Quoted values.
+   * @param delimiter          delimiter for the file
+   * @return Array of all the column values within the provided row.
+   */
+  public static String[] getRowValues(String rawLine, boolean enableQuotedValues, String delimiter) {
+    if (enableQuotedValues) {
```
albertshau: nit: in these situations, can remove some nesting and improve readability by checking the inverse:

```
if (!enableQuotedValues) {
  return ...
}
...
```

albertshau (approved):

PR #193, opened 2025-04-15
Update cdap version to 6.11.0 in pom.xml

psainics:
Update CDAP version to 6.11.0 in pom.xml, as snapshot is not available in repo causing build failures

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -76,8 +76,8 @@
 
   <properties>
     <awaitility.version>3.1.6</awaitility.version>
-    <cdap.version>6.11.0-SNAPSHOT</cdap.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -76,7 +76,7 @@
 
   <properties>
     <awaitility.version>3.1.6</awaitility.version>
-    <cdap.version>6.11.0-SNAPSHOT</cdap.version>
+    <cdap.version>6.11.0</cdap.version>
     <cdap.plugin.version>2.13.0-SNAPSHOT</cdap.plugin.version>
```
itsankit-google: similar change here, please use `2.13.0`

psainics: Updated !

itsankit-google (approved):

PR #200, opened 2026-01-27
[PLUGIN-1941] Set correct ssl factory when using OAUTH

psainics:
## Set correct ssl factory when using OAUTH

Jira : [PLUGIN-1941](https://cdap.atlassian.net/browse/PLUGIN-1941)

### Description

- adds correct ssl factory when using oauth

### Bug
HTTP Plugin does not skip SSL cert validation when using OAUTH

### Steps to recreate
1. Use any https endpoint with self signed SSL cert
2. Turn off SSL cert validation
3. Try to request with any auth (say basic auth) and then with OAUTH

OAuth will fail will cert errors

[PLUGIN-1941]: https://cdap.atlassian.net/browse/PLUGIN-1941?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/io/cdap/plugin/http/common/http/OAuthUtil.java
+++ b/src/main/java/io/cdap/plugin/http/common/http/OAuthUtil.java
@@ -83,6 +83,13 @@ public class OAuthUtil {
         // get accessToken from service account
         return OAuthUtil.getAccessTokenByServiceAccount(config);
       case OAUTH2:
+        if (config instanceof BaseHttpSourceConfig) {
+          try (CloseableHttpClient client = HttpClients.custom()
+              .setSSLSocketFactory(new SSLConnectionSocketFactoryCreator((BaseHttpSourceConfig) config).create())
+              .build()) {
+            return getAccessToken(client, config);
+          }
+        }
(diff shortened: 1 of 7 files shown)


src/main/java/io/cdap/plugin/http/common/http/OAuthUtil.java
```diff
@@ -83,6 +83,13 @@ public static AccessToken getAccessToken(BaseHttpConfig config) throws IOExcepti
         // get accessToken from service account
         return OAuthUtil.getAccessTokenByServiceAccount(config);
       case OAUTH2:
+        if (config instanceof BaseHttpSourceConfig) {
```
itsankit-google, line 86: why only in source and not in sink?

psainics: There is no `SSL/TLS` config section in http sink

itsankit-google (approved):

PR #205, opened 2026-04-20
[PLUGIN-1956] Added a fix for OAuth2 Proxy Routing

psainics:
## Added a fix for OAuth2 Proxy Routing

Jira : [PLUGIN-1956](https://cdap.atlassian.net/browse/PLUGIN-1956)

### Description

When the HTTP plugin (source or sink) is configured with OAuth2 authentication and a proxy (proxyUrl, proxyUsername, proxyPassword), the request that fetches the OAuth2 access token ignores the proxy settings. It is sent directly to the token endpoint, which causes an UnknownHostException in restricted environments (like Cloud Data Fusion / Dataproc) where only the proxy has outbound network access.
Root Cause
In OAuthUtil.getAccessToken(BaseHttpConfig config), the code builds its own HTTP client but never applies the proxy settings from the config. Currently, it uses HttpClients.custom() (for the source path) or HttpClients.createDefault() (for the sink path), completely bypassing setProxy(...) and setDefaultCredentialsProvider(...).

### The Fix
Refactored the OAUTH2 case in OAuthUtil.java to use a unified HttpClientBuilder. The builder now:
Conditionally applies the SSLConnectionSocketFactory (if it's a BaseHttpSourceConfig).
Extracts the proxy settings from the config object and applies the HttpHost and BasicCredentialsProvider (if configured) before calling .build().

### Proof / Verifications

- Pipeline execution logs showing the UnknownHostException failing to resolve the token URL in a restricted environment prior to this fix.
i1

- Local mock proxy server logs confirming that the POST /token request is now successfully intercepted and routed through the configured proxy port.

i2

i3

i4

[PLUGIN-1956]: https://cdap.atlassian.net/browse/PLUGIN-1956?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/io/cdap/plugin/http/common/http/OAuthUtil.java
+++ b/src/main/java/io/cdap/plugin/http/common/http/OAuthUtil.java
@@ -24,11 +24,17 @@ import io.cdap.plugin.http.common.BaseHttpConfig;
 import io.cdap.plugin.http.common.OAuth2GrantType;
 import io.cdap.plugin.http.common.pagination.page.JSONUtil;
 import io.cdap.plugin.http.source.common.BaseHttpSourceConfig;
+import org.apache.http.HttpHost;
+import org.apache.http.auth.AuthScope;
+import org.apache.http.auth.UsernamePasswordCredentials;
+import org.apache.http.client.CredentialsProvider;
 import org.apache.http.client.entity.UrlEncodedFormEntity;
 import org.apache.http.client.methods.CloseableHttpResponse;
 import org.apache.http.client.methods.HttpPost;
 import org.apache.http.client.utils.URIBuilder;
+import org.apache.http.impl.client.BasicCredentialsProvider;
 import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.impl.client.HttpClientBuilder;
 import org.apache.http.impl.client.HttpClients;
 import org.apache.http.message.BasicHeader;
 import org.apache.http.message.BasicNameValuePair;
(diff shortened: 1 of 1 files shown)


src/main/java/io/cdap/plugin/http/common/http/OAuthUtil.java
```diff
@@ -83,14 +89,26 @@ public static AccessToken getAccessToken(BaseHttpConfig config) throws IOExcepti
...
-            return getAccessToken(client, config);
+          httpClientBuilder.setSSLSocketFactory(
+                  new SSLConnectionSocketFactoryCreator((BaseHttpSourceConfig) config).create()
+          );
+        }
+        // Apply Proxy settings if they exist
+        if (!Strings.isNullOrEmpty(config.getProxyUrl())) {
+          HttpHost proxyHost = HttpHost.create(config.getProxyUrl());
+
+          if (!Strings.isNullOrEmpty(config.getProxyUsername()) && !Strings.isNullOrEmpty(config.getProxyPassword())) {
```
itsankit-google, line 103: should the else part throw an error?

Do we have validation in place earlier which checks that these cannot be empty when `proxyUrl` is specified?

vishwasvaidya-cloudsufi: Using the ProxyURL without Credentials is a valid use case, so adding an else part for the same is not required. 
Similar type of validations are present where the proxy configuration is implemented. I have also tested these changes locally to verify the same.

itsankit-google: I see, sounds good.

itsankit-google (approved):

